PR #8, opened 2015-11-18
Include `node.text` when helpful

seanpdoyle:
Without specifying `node.text`, error messages might not always contain
helpful information.

For example, the following was raised for an `<a>` tag without an `href`:

```
A11yError:  has a non-meaningful href, it should point to an actual
link.
```

When `${node}` replaced with `${node.text}`, the following was produced:

```
A11yError: "Waiting for a submission..." has a non-meaningful href, it
should point to an actual link.
```

### DIFF
--- a/test-support/helpers/a11y/helpers/actions.js
+++ b/test-support/helpers/a11y/helpers/actions.js
@@ -26,7 +26,7 @@ if (!Element.prototype.matches) {
     let el = this;
     let matches = (el.document || el.ownerDocument).querySelectorAll(selector);
     let i = 0;
-    
+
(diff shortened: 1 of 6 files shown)


seanpdoyle:
Maybe `text` might not be the best idea, but `node.toString()` is a little too unreliable.

Maybe `node.outerHTML` or `node.innerHTML`

trentmwillis:
Conflicts due to the other error PR I just merged. Not sure if this is needed any more since we log the object now.

seanpdoyle:
closing in favor of https://github.com/trentmwillis/ember-a11y-testing/pull/11

PR #9, opened 2015-11-18
Accept `<label><input></label>`

seanpdoyle:
From the [Google Chrome accessibility developer tools][chrome]:

```html
<!-- Good: using label-wrapped -->
<!-- In this case both "Address:" and the contents of the text field will be read. -->
<label>
  Address:
  <input>
</label>
```

[chrome]: https://github.com/GoogleChrome/accessibility-developer-tools/wiki/Audit-Rules#controls-and-media-elements-should-have-labels


trentmwillis:
:+1:

PR #10, opened 2015-11-18
Allow `<form>` elements to have `{{action}}`

seanpdoyle:
Form elements can have [actions triggered by submit][submit]:

```hbs
<form {{action "save" on="submit"}}`>
</form>
```

[submit]: https://github.com/emberjs/ember.js/blob/1a2cd166a86219717a0cde1df4dab497d8dda3e3/packages/ember-routing-htmlbars/lib/keywords/element-action.js#L51

### DIFF
--- a/test-support/helpers/a11y/helpers/actions.js
+++ b/test-support/helpers/a11y/helpers/actions.js
@@ -14,11 +14,17 @@ const FOCUS_SELECTORS = [
   'a[href]',
   'area[href]',
   'iframe',
-  '[tabindex]:not([tabindex="-1"])'
+  '[tabindex]:not([tabindex="-1"])',
+];
+
+const FOCUS_EXCEPTIONS = [
+  'form',
 ];
 
 // Selector to find any Ember-bound actions
 const ACTION_SELECTOR = '[data-ember-action]';
+
+const MATCHES = (el) => (selector) => el.matches(selector);
 
 // Add polyfill for browsers that don't have Element.matches (including Phantom)
 if (!Element.prototype.matches) {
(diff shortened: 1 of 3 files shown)


test-support/helpers/a11y/helpers/actions.js
```diff
@@ -14,7 +14,8 @@ const FOCUS_SELECTORS = [
   'a[href]',
   'area[href]',
   'iframe',
-  '[tabindex]:not([tabindex="-1"])'
+  '[tabindex]:not([tabindex="-1"])',
+  'form'
```
seanpdoyle: @trentmwillis adding `'form'` to `FOCUS_SELECTORS` might not be the best approach. 

Since the functions declared in this file are part of the package's public API, I'm not sure of a better way.

trentmwillis: I think introducing another array to check against would be the best approach.

Maybe something like `FOCUS_EXCEPTIONS`. Then the logic can be make sure it is focusable OR that it is one of the exceptions to that rule.

test-support/helpers/a11y/helpers/actions.js
```diff
@@ -43,7 +49,8 @@ if (!Element.prototype.matches) {
  * @return {Boolean|Error}
  */
 export function actionIsFocusable(app, el) {
-  if (!FOCUS_SELECTORS.filter((selector) => el.matches(selector)).length) {
+  if (!FOCUS_SELECTORS.filter(MATCHES(el)).length &&
```
seanpdoyle: @trentmwillis this is ready for another look.

seanpdoyle:
@trentmwillis if we merge this, could you cut a new release?

trentmwillis:
Merging and will cut a new release shortly. Thanks for the contributions!

Issue #2, opened 2015-06-17
Better logging of errors


trentmwillis:
Currently all errors just inject any offending HTMLElement into the error message string. This makes identifying the exact offending element difficult. Instead, we should switch to a model that logs the element separately so it can be inspected and identified.

majew7:
Was the fix removed?

This bug fix (9505490a9), made understanding the offending element easier by improving the `A11yError` constructor function.  However, I see that e599bccbdc36940, named _"Remove old approach"_ from May 18, 2016, seems to have removed the improvement that this issue ticket was all about.

**Question)** What does "Remove old approach." mean?

Understanding the results is difficult when the console is throwing multiple errors, because I don't know the offending HTML element.

PR #11, opened 2015-11-18
Log `A11yError` element

seanpdoyle:
Closes [#2].

When throwing an `A11yError`, include the element separately from the
message.

Then, during `a11yTest` invocation, catch the error, log it to the
console, then rethrow.

Ensure the error [preserves the current frame's stack trace][mozilla].

[#2]: https://github.com/trentmwillis/ember-a11y-testing/issues/2
[mozilla]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#Custom_Error_Types


trentmwillis:
Thanks! I appreciate the error improvements!

PR #12, opened 2015-12-09
Do not throw error if alt tag is present but empty

mydea:
Right now, if you have an empty alt-attribute, an error is thrown:

```html
<img src="myimage.jpg" alt="">
```

However, elements are actually allowed to have an empty alt-attribute if they are truly decorative:
http://www.w3.org/TR/WCAG20-TECHS/H67.html

Note the difference between having an empty alt-attribute and no alt-attribute at all.

I changed the check from `getAttribute("alt")` to `hasAttribute("alt")`. I also adapted the tests to reflect this behaviour and added a new test for this use case.


trentmwillis:
Sounds good, thanks! Merging, will cut a new release later today.

PR #16, opened 2016-05-18
Fix typo in combobox for aria role autocomplete

binhums:
Was looking for the way we import axe-core, was surprised to see a copied file here.

https://embercommunity.slack.com/archives/topic-a11y/p1463595296001123


nathanhammond:
Issue was relevant to the older version of ember-a11y-testing.

Issue #5, opened 2015-06-17
Visual feedback for issues during development


trentmwillis:
After making components aware of accessibility, they should be able to manipulate the visual appearance of an offending element during development to make it obvious something is wrong (turn it red? Big arrow pointing to it? Open to suggestions).

noopkat:
Kahn Academy's [Tota11y](https://khan.github.io/tota11y/) tool does a tremendous job at doing this. I wonder if it could be forked and extended for this test suite's use to include things like elements with an ember action not being focussable etc? It's MIT license which would allow this :smile:

trentmwillis:
This has been implemented (more-or-less) via the violation warning level styles.

PR #20, opened 2016-05-19
Ember 2.x upgrades + fixes

nathanhammond:
Hello! 👋 

Thank you for writing this add-on! We're looking to use this in both our dev and CI flow here at work, and are deprecating [ember-a11y-testing](https://github.com/trentmwillis/ember-a11y-testing) as advised.

**I'm opening this PR as an early WIP / RFC.**

I thought now would be a good opportunity to update the dependencies on this add-on so we can have clean slate for regular maintenance going forward (which I'm happy to volunteer with).

I also rolled in the fragment fix from #5 in as I was running into the same error with fragments in our stack.

Travis config has added newer Node versions added to ensure everything is passing there.

Lastly, there was a bug we were experiencing to do with red border highlighting the violating ember component on the page. The DOM selector string that axe-core returns as the offending element includes the Ember component's id, eg. `"#ember727 > .bundle__see-more-content > .bundle__see-more-label"`. Using the component's `$.()` method makes it impossible to include itself while hunting for the exact DOM element to highlight. Let me know if I'm missing something obvious though!

There's a performance hit for the fix I implemented, as I wasn't as comfortable with something more fragile such as regex'ing out the component's id from the axe-core `node.target` selector string.

### RFC

+ I'm new to Ember add-on architecture. Am I doing this right?
+ Will these changes cause any backwards compatibility issues with Ember pre 2.x?
+ Should I add multiple Ember versions to the Ember Try config?

### todo

- [x] get the styles working again for the red border on violations in the UI (dev flow)
- [x] Remove the version bump.
- [x] Add back in the keywords.
- [x] Add ember-lts to the Ember Try config.
- [x] Include ember-lts and 1.13 into the `travis.yml` file.

Ported from https://github.com/trentmwillis/ember-axe/pull/6 ... @noopkat you have access to my fork.


nathanhammond:
LGTM!

PR #22, opened 2016-05-22
Allow passing component configuration options in environment configuration

BrianSipple:
I like the additional options being enabled for components, but I've found that sometimes it can be useful to A) set app-wide component defaults, and/or B) easily control these app-wide settings on the fly during development.

This led me to making  `ember-a11y-testing` configurable through the app’s ENV hash:

```javascript
ENV['ember-a11y-testing'] = {
    componentOptions: {
      turnAuditOff: true
      ...
    }
  }
};
```

All we’d have to do is have our instance initializer read in the hash from `../config/environment`, and the component properties for `axeOptions`, `axeCallback`, `turnAuditOff` have the opportunity to be set to what’s defined on `ENV['ember-a11y-testing'].componentOptions.` (Side note: I layered in the `componentOptions` property so that -- if we're okay with this approach -- we could flexibly add similarly separated hashes for whatever `xOptions` might be desirable in the future.)

Additionally, trying to update `README.md` for this lead me to finding a number of ways that it could be structured more cleanly — and I ended up moving a few pieces around in ways that I think help the overall readability (regardless of the other changes introduced by the PR).

### DIFF
--- a/app/instance-initializers/axe-component.js
+++ b/app/instance-initializers/axe-component.js
@@ -1,3 +1,5 @@
+import ENV from '../config/environment';
+
 /**
  * Variable to ensure that the initializer is only ran once. Though in this
  * particular case, running more than once shouldn't cause side-effects.
@@ -8,27 +10,32 @@ let hasRan = false;
 export function initialize(application) {
   if (hasRan) { return; }
 
+  const appConfig = ENV['ember-a11y-testing'] || {};
+  const { componentOptions: { axeOptions, turnAuditOff, axeCallback } = {} } = appConfig;
+
   Ember.Component.reopen({
     /**
      * An optional callback to process the results from the a11yCheck.
      * @public
      * @type {Function}
      */
-    axeCallback: undefined,
+    axeCallback,
 
     /**
      * An optional options object to be used in a11yCheck.
      * @public
      * @type {Object}
      */
-    axeOptions: undefined,
+    axeOptions,
 
     /**
      * Turns off the accessibility audit during rendering.
+     * Defaults to false if not set in the application's configuration.
+     *
      * @public
      * @type {Boolean}
      */
-    turnAuditOff: false,
+    turnAuditOff: typeof turnAuditOff === 'boolean' ? turnAuditOff : false,
 
     /**
      * Runs an accessibility audit on any render of the component.
--- a/app/instance-initializers/axe-component.js
+++ b/app/instance-initializers/axe-component.js
@@ -10,12 +10,14 @@ let hasRan = false;
 export function initialize(application) {
   if (hasRan) { return; }
 
-  const appConfig = ENV['ember-a11y-testing'] || {};
-  const { componentOptions: { axeOptions, turnAuditOff, axeCallback } = {} } = appConfig;
+  const addonConfig = ENV['ember-a11y-testing'] || {};
+  const { componentOptions: { axeOptions, axeCallback } = {} } = addonConfig;
 
   Ember.Component.reopen({
     /**
      * An optional callback to process the results from the a11yCheck.
+     * Defaults to `undefined` if not set in the application's configuration.
+     *
      * @public
      * @type {Function}
      */
@@ -23,6 +25,7 @@ export function initialize(application) {
 
     /**
      * An optional options object to be used in a11yCheck.
+     * Defaults to `undefined` if not set in the application's configuration.
      * @public
      * @type {Object}
      */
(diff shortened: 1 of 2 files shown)


app/instance-initializers/axe-component.js
```diff
@@ -8,27 +10,32 @@ let hasRan = false;
 export function initialize(application) {
   if (hasRan) { return; }
 
+  const appConfig = ENV['ember-a11y-testing'] || {};
```
nathanhammond: Let's call it `addonConfig`.

app/instance-initializers/axe-component.js
```diff
@@ -8,27 +10,32 @@ let hasRan = false;
...
 
     /**
      * Turns off the accessibility audit during rendering.
+     * Defaults to false if not set in the application's configuration.
+     *
      * @public
      * @type {Boolean}
      */
-    turnAuditOff: false,
+    turnAuditOff: typeof turnAuditOff === 'boolean' ? turnAuditOff : false,
```
nathanhammond: I'm ~99% sure that we don't want to allow this to be turned off globally. Having one flag which makes most of the work go away will have the same consequence as our CSP addon: "fine, we'll just open it up to `unsafe-inline`!" ... "fine! we'll just turn it off!"

app/instance-initializers/axe-component.js
```diff
@@ -8,27 +10,32 @@ let hasRan = false;
 export function initialize(application) {
   if (hasRan) { return; }
 
+  const appConfig = ENV['ember-a11y-testing'] || {};
+  const { componentOptions: { axeOptions, turnAuditOff, axeCallback } = {} } = appConfig;
```
nathanhammond: The destructuring here is great, but it makes it less clear what's going on.

nathanhammond:
Thoughts:
- I like the planning for the future with the component-specific options sub-hash. This makes a lot of sense from a configuration perspective. Makes it less likely that we have to do major version bumps in the future.
- If we remove the global config for disabling I'm in favor of landing.
- Not sure why Travis is unhappy.

So:
- [ ] Second review.
- [ ] Remove global disabling.
- [ ] Make Travis happy.

app/instance-initializers/axe-component.js
BrianSipple: That’s a REALLY interesting (and good) point. 

My approach here was considering a similar reaction on the opposite end of the spectrum: someone introducing `ember-a11y-testing` into existing application, and all of a sudden having a hundred components being audited in ways they haven’t yet had a chance to control.

So I'm split -- but largely on your side now 😃. I can totally get behind the notion of not having a global switch, and perhaps exploring more config options if we get more user feedback around the matter in the future.

BrianSipple:
RE: Travis -- it's 😒 about master, too. I'm looking into it now, and I'll try to check something in with a separate PR that will get us green here.

app/instance-initializers/axe-component.js
nathanhammond: The configuration of which rules you enable inside of aXe is probably the way to go for this sort of behavior instead of turning off testing for all components. Turning on any linting or automated testing approach in an existing project is hard (and I've experienced that pain) and I don't mind the "surprise, you've got a lot of work to do!" because I don't think it's particularly surprising.

I don't want to have to set up a linter on top of this which checks to make sure you're not doing the wrong thing with your project config.

nathanhammond: The configuration of which rules you enable inside of aXe is probably the way to go for this sort of behavior instead of turning off testing for all components. Turning on any linting or automated testing approach in an existing project is hard (and I've experienced that pain) and I don't mind the "surprise, you've got a lot of work to do!" because I don't think it's particularly surprising.

I don't want to have to set up a linter on top of this which checks to make sure you're not doing the wrong thing with your project config.

noopkat:
I like this idea a lot! 👏 
+1 about not having an 'easy off' global switch.

BrianSipple:
The global switch has been removed, Travis is happy, I'm happy -- I'd say this is just about ready to go.

nathanhammond:
@BrianSipple Can you grant me access to your fork? I've fiddled with git things a bit. For PRs we'd rather rebase onto master than merge master into them–especially when the changes are disjoint sets.

BrianSipple:
@nathanhammond Ah, I was planning on rebasing before initiating the actual merge -- is that what you're referring to?

nathanhammond:
Yep yep! This PR should just be the two commits. Rebase and this is good to land!

PR #24, opened 2016-05-23
chore(editorconfig): removed redundant rules

sarbbottam:
```
indent_style = space
indent_size = 2
```
are already part of `*`, no need to specify them explicitly for `*.js`, `*.hbs`, `*.css` and `*.html`.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,7 +4,6 @@
 
 root = true
 
-
 [*]
 end_of_line = lf
 charset = utf-8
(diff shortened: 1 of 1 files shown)


nathanhammond:
Hey @sarbbottam! We're maintaining this to [match the default ember-cli app blueprint](https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/.editorconfig) so that we don't have unnecessary noise during upgrades. You're correct in that they're duplicate, I'd open a PR against ember-cli. We'll maintain this to match the default ember-cli blueprint so if a change lands there it will land here.

Closing in favor of upstreaming which will eventually land here. Thanks for submitting a PR!

sarbbottam:
no worries, raised https://github.com/ember-cli/ember-cli/pull/5924 against `ember-cli`

PR #25, opened 2016-06-02
Add more descriptive violation logging

aliciasedlock:
Original PR from ember-axe: https://github.com/trentmwillis/ember-axe/pull/7

As we are about to use this add-on in a new application, the other developer and I thought it would be nice to have more descriptive logging in the console rather than sifting through the `violations` object output.

This both updates the default messages that is displayed inside the browser test runner, along with more precise console logging, rather than dumping the entire `violations` object flat out.

#### New test runner messaging
screen shot 2016-05-12 at 8 34 10 pm

#### New console level information
screen shot 2016-06-09 at 11 08 32 am

### DIFF
--- a/content-for/test-body-footer.html
+++ b/content-for/test-body-footer.html
@@ -14,12 +14,33 @@
     a11yCheckCallback: function(results) {
       var violations = results.violations;
 
-      for (var i = 0, l = violations.length; i < l; i++) {
-        Ember.Logger.error('Violation #' + (i + 1), violations[i]);
-      }
+      if (violations.length) {
+        Ember.Logger.error('ACCESSIBILITY VIOLATIONS: ' + violations.length);
 
-      Ember.assert('The page should have no accessibility violations.', !violations.length);
+        for (var i = 0, l = violations.length; i < l; i++) {
+          var violation = violations[i];
+          var violationNodeString = axe.ember.nodesToString(violation.nodes);
+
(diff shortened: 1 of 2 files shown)


content-for/test-body-footer.html
```diff
@@ -14,14 +14,35 @@
...
     /**
+     * Takes the violations' nodes param and converts it to
+     * a string that can be displayed with the console error.
+     * @param {Array} nodes
+     * @return {String}
+     */
+     nodesToString: function(nodes) {
+       return nodes.map(function(node) {
+         return node.html;
+       }).join(', ');
```
nathanhammond: What if this string is absurdly long?

aliciasedlock: Good point. There is potential for it to contain a lot of nodes, so perhaps the intent of having the node list at this level makes it less useful.

I can potentially log how many specific violations of that type are present and instead provide an object or array of the offending elements.

aliciasedlock:
Change has been implemented locally, but I would like to test this on our application prior to marking this as ready.

nathanhammond:
Remove the WIP flag and drop another note here when you're calling it ready and I'll merge. 😄 Thanks for your efforts on this!

aliciasedlock:
@nathanhammond Second pass done! Screenshot is updated, we now list the number of violations and a separate violations array with the specific nodes in it, rather than converting them to a string.

nathanhammond:
Thank you @aliciasedlock!

PR #31, opened 2016-06-21
Set turnAuditOff in init and clean up Ember.testing toggling in tests

BrianSipple:
After playing around in the current test suite a bit, it seems like we could stand to clean up both the way we set `turnAuditOff`, and the way we then test for it.

1) Rather than inspecting `Ember.testing` every time the `didRender` hook is called, we could just check it once inside of `init`. This way, the logic inside of `didRender` doesn't have to worry about `Ember.testing` and can use  `turnAuditOff` off as the "sole source of truth”.

2) It also seems like we only need to test for this behavior once -- and so I remove the `Ember.testing` toggling within the `audit is not run on didRender when in testing mode` test

**[NOTE]** The current CI failures here are resolved in https://github.com/ember-a11y/ember-a11y-testing/pull/36

### DIFF
--- a/app/instance-initializers/axe-component.js
+++ b/app/instance-initializers/axe-component.js
@@ -7,11 +7,12 @@ import ENV from '../config/environment';
  */
 let hasRan = false;
 
-export function initialize(application) {
-  if (hasRan) { return; }
+export function initialize(application, opts = { forceRun: false }) {
+  if (hasRan && !opts.forceRun) { return; }
 
   const addonConfig = ENV['ember-a11y-testing'] || {};
   const { componentOptions: { axeOptions, axeCallback } = {} } = addonConfig;
+
 
   Ember.Component.reopen({
     /**
@@ -34,10 +35,11 @@ export function initialize(application) {
     /**
      * Turns off the accessibility audit during rendering.
      *
+     * defaults to true in testing, false otherwise
      * @public
      * @type {Boolean}
      */
-    turnAuditOff: false,
+    turnAuditOff: !!Ember.testing,
 
     /**
      * An array of classNames to add to the component when a violation occurs.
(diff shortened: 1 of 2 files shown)


app/instance-initializers/axe-component.js
```diff
@@ -34,21 +34,27 @@ export function initialize(application) {
     /**
      * Turns off the accessibility audit during rendering.
      *
+     * defaults to true in testing, false otherwise
      * @public
      * @type {Boolean}
      */
     turnAuditOff: false,
```
nathanhammond: Since we're reopening here we could literally just set it the first time on the prototype? `turnAuditOff: !Ember.testing`

This is something that should never change at runtime so that should be safe.

tests/unit/instance-initializers/axe-component-test.js
```diff
@@ -40,12 +40,14 @@ test('initializer should not re-open Ember.Component more than once', function(a
 test('audit is run on didRender when not in testing mode', function(assert) {
   initialize(application);
 
+  // In order for the audit to run, Ember.testing has to be false
+  // when the component is initialized
+  Ember.testing = false;
   let component = Ember.Component.create({});
+  Ember.testing = true;
```
nathanhammond: That of course makes tests like this a touch harder to write, but oh well–everything else is much faster.

BrianSipple:
@nathanhammond [I think this is a good suggestion](https://github.com/ember-a11y/ember-a11y-testing/pull/31#discussion_r67803511). The only trouble I'm running into is that the entire exported `initialize` function has closure over a `hasRan` variable, which it checks before reopening the component.

This means that even toggling `Ember.testing` within the tests -- before or after calling `initialize(application)` -- has no way to impact the initialization of the prototype that happens in `reopen`. (Interestingly, this also means that our [first test](https://github.com/ember-a11y/ember-a11y-testing/pull/31/files#diff-3c6819060e23ca79ba98f105186e43ebR37) only succeeds because it's, well, first. We might want to denote that somewhere 😄).

Because of this this, I pushed up a solution that gives the initialize function a second `opts` parameter — which allows the setting of a `forceRun` property that defaults to false. (The config-object based approach lets us be explicit about what we're doing when using it: `initialize(application, { forceRun: true });`.)

What's getting in getting in the way now is a curious behavior of `didRender` that I'm noticing.
 
**On every other test suite run**, starting with the first, the component's `didRender` hook is being called twice in, [this test](https://github.com/ember-a11y/ember-a11y-testing/pull/31/files#diff-3c6819060e23ca79ba98f105186e43ebR43), when a component is [appended to the DOM](https://github.com/ember-a11y/ember-a11y-testing/pull/31/files#diff-3c6819060e23ca79ba98f105186e43ebR58) and twice again [when we manually trigger it](https://github.com/ember-a11y/ember-a11y-testing/pull/31/files#diff-3c6819060e23ca79ba98f105186e43ebR63). On every even-numbered run of the tests, `didRender` only called once per action, and everything passes as expected. 

I put in some logging statements that can let you easily see what's going on in the console after running `ember test --serve` -- and then pressing `return` to rerun. If anyone on @ember-a11y/core has seen this before, or has any idea as to what the cause might be, I'm quite fascinated by it at the moment 😄.

MelSumner:
Closing this based on amount of time that has passed w/o action- please feel free to resubmit the PR once everything is working, though! Thank you so much :)

PR #32, opened 2016-06-21
Use Chrome for CI tests

BrianSipple:
From what I've been able to gather, the test failures in https://github.com/ember-a11y/ember-a11y-testing/pull/28 all seem to be due to PhantomJS not supporting either Array.find or HTMLElement.remove(). 

All tests pass in Chrome.

With that in mind, now might be a great opportunity to set up Chrome for CI tests -- especially given how much we're doing with the DOM. All we need is a few `travis.yml` and `testem` tweaks, and we'd be all set.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,15 @@ node_js:
   - "5"
   - "6"
 
-sudo: false
+sudo: required
+dist: trusty
+
+addons:
+  apt:
+    sources:
+    - google-chrome
+    packages:
+    - google-chrome-stable
 
 cache:
   directories:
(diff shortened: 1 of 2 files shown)


nathanhammond:
I'm actually opposed to this since every Ember app is set up to use Phantom by default. It'd be really easy for us to accidentally introduce a feature that didn't work on Phantom once we move CI to Chrome. I'd much rather ship a polyfill for this behavior in our tests so that we don't accidentally introduce regressions–especially since our code is designed to be run at test time.

BrianSipple:
@nathanhammond Good point, I hadn't yet wrapped my brain around the fact that we're designing for use in tests here as well.  I'll close the issue -- I think those concerns far outweigh the benefits atm -- but hey, it's good to know now that this challenge might come up!.

Issue #18, opened 2016-05-19
Improve logging and debugging of violations


nathanhammond:
Initial implementation of logging for the audits has been minimal. Currently, it just outputs the violations in numeric order follow by a dump of the object. Example:

![screen shot 2015-06-25 at 10 10 41 pm](https://cloud.githubusercontent.com/assets/3526753/8371154/fee50ac8-1b87-11e5-94e7-34831424965e.png)

This isn't the worst, but can definitely be made better before the official 1.0 release. Following are proposed features for better logging of violations (subject to change):

**General Improvements**
- [x] Present information in an easy-to-digest format before dumping the object.
  - Potential format: `[impact]: [help]. Offending markup is: [nodes.html].`
  - Ex: `CRITICAL: Form elements must have labels. Offending markup is: <input id="test" type="test">.`

**Component Improvements**
- [ ] Include the Component name/reference (e.g., `component.toString()`) so that it is clear who the offender was
  - [ ] Provide hook to break on a specific violation

**Acceptance Test Improvements**
- [x] For Acceptance Test audits, specify the run number of the audit
  - [x] Provide hook to break on a specified run number to aid in debugging

ksin:
Not sure if there's already a way to do this in development, but perhaps an object that keeps track of all violations and can be accessed in the console? Ideally, it updates as a violation appears. 

ie: After submitting a form, an error message appears where the error message triggers a violation. The violations tracker is then updated with the violation or reference of the violation. The dev can call something like `violations.count` or `violations.lastObject` to know of any state changes to the tracker at any time.

My main desire for this is that when I see a ton of violations on a page, I feel a bit overwhelmed seeing something like:

![screen shot 2016-07-08 at 1 56 08 am](https://cloud.githubusercontent.com/assets/6626989/16678657/31340a96-44af-11e6-8581-84920997ac8d.png)

-- especially as a first-time user of the addon and newbie to a11y concepts. I feel like a simple API like this would make seeing a ton of violations on a page at once a little easier to manage.

Anyway, I'm sure the improvements listed above will help a lot, but just wanted to throw my suggestion in the box! If there's an easy way to get something like this up and you can point me in the right direction, I'd be happy to work on a quick PR.

(I know it's my own fault for letting the violations get so bad... :sob:)

nathanhammond:
@ksin Hokay, that's a complete and total mess, I agree. Would you be interested in creating a util object for this addon that has a few helper methods?

[...]

PR #35, opened 2016-07-12
Expose violationsHelper to global scope.

ksin:
Util object that exposes some helper methods in the console. Hopefully makes dealing with violations in development a little less messy.

As discussed @nathanhammond 
https://github.com/ember-a11y/ember-a11y-testing/issues/18#issuecomment-231880458

### DIFF
--- a/addon/utils/violations-helper.js
+++ b/addon/utils/violations-helper.js
@@ -1,0 +1,103 @@
+import Ember from 'ember';
+/**
+ * @module ember-a11y-testing
+ *
+ *
+ * Use the violations helper to inspect the collection of violations.
+ * This provides a set of handy helper methods so you can quickly query
+ * and filter the violations from the console.
+ *
+ *
+ * Helper methods include:
+ * - count
+ * - first
+ * - last
+ * - push
+ * - filterBy
+ *
+ * @public
+ * @class ViolationsHelper
+ */
+export default class ViolationsHelper {
+  /**
+   * Instantiate by calling either:
+   *
+   * new ViolationsHelper() or
+   * new ViolationsHelper(violation1, violation2, ...)
+   *
+   */
+  constructor() {
+    this.violations = Array.prototype.slice.call(arguments);
+    this.hasLoggedTip = false;
+  }
+
+  /**
+   * Alias method to return the current number of violations
+   *
+   * @public
+   * @type {Number}
+   */
+  get count() { return this.violations.length; }
+
+  /**
+   * Alias method to return the first violation
+   *
+   * @public
+   * @type {Object}
+   */
+  get first() { return this.violations[0]; }
+
+  /**
+   * Alias method to return the last violation
+   *
+   * @public
+   * @type {Object}
+   */
+  get last() { return this.violations[this.count - 1]; }
+
+  /**
+   * Alias method to push a violation into the collection
+   *
+   * @public
+   * @return {Void}
+   */
+  push(violation) {
+    this.violations.push(violation);
+  }
+
+  /**
+   * Filters violations by a key value pair such as:
+   * key = "impact", value = "critical"
+   *
+   * Special cases:
+   * The "rule" key is an alias for the "id" key.
+   * The "node" key filters violations by nodes with the selector of value.
+   *
+   * @public
+   * @return {Array}
+   */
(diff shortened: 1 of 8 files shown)


addon/utils/violations-helper.js
```diff
@@ -0,0 +1,103 @@
...
+ * - count
+ * - first
+ * - last
+ * - push
+ * - filterBy
+ *
+ * @public
+ * @class ViolationsHelper
+ */
+export default class ViolationsHelper {
```
Robdel12: 😍  love me some ES6 classes

app/instance-initializers/violations-helper.js
```diff
@@ -0,0 +1,11 @@
+import ViolationsHelper from 'ember-a11y-testing/utils/violations-helper';
+
+export function initialize(application) {
+  window.violationsHelper = new ViolationsHelper();
```
Robdel12: Do we really want to expose this globally on the window object?

ksin: The idea is so the dev can query it in the console, so setting it globally on window was what came to mind first. Do you think there's a better place?

MelSumner: @Robdel12 will we take a serious performance hit?

ksin: I don't believe so. Violations are pushed into the helper during audit:
https://github.com/ksin/ember-a11y-testing/blob/62c81262c68a635f953d2a0a0795f417c9bfdcc9/app/instance-initializers/axe-component.js#L87

Robdel12: No serious performance hits, but exposing globals is something you want to be very sure about :)

Robdel12: What if we did something like [`ember-devtools`](https://github.com/aexmachina/ember-devtools#globalize) did and only set it to a global if you want to?

ksin: @Robdel12 That sounds good to me! Do you think this "feature" warrants a note in the README?

Robdel12: Probably. :) 

My concern is exposing the global in production builds, which might not happen. But that's what I'm trying to guard against

ksin: Yeah, totally! Huge oversight I didn't guard against that. >_<

ksin: @Robdel12 instead of guarding all the places `violationsHelper` is mentioned, how do you feel about just excluding the initializer in production like: https://github.com/ember-a11y/ember-a11y-testing/blob/master/index.js#L55

Treat it like a public API in development/testing. No fiddling with ENV.

Robdel12: I like it!

ksin:
☝️  amended with changes.

tests/acceptance/violations-test.js
```diff
@@ -41,3 +41,28 @@ test('marking DOM nodes with violations', function(assert) {
...
+    logTipSpy = sandbox.spy(window.violationsHelper, 'logTip');
+  });
+
+  andThen(() => {
+    assert.equal(window.violationsHelper.count, 2, "Two violations are found in the violationsHelper");
+
+    assert.ok(logTipSpy.calledOnce, "logTip is only called once after all components are rendered and violations logged");
+
+    // Turn testing mode back on to ensure validity of other tests
+    Ember.run(function() { Ember.testing = true; });
```
sarenji: if either of these asserts fail, will this line be reached?

app/instance-initializers/violations-helper.js
sarenji: Could also consider `fastboot` and perhaps exclude it from those builds as well

tests/acceptance/violations-test.js
ksin: Yep! In fact, wrapping this in `Ember.run` is a little excessive.

app/instance-initializers/violations-helper.js
ksin: That's a good idea! But perhaps needs a little more discussion / out of scope for this PR?

Robdel12:
This looks good to me, ping @ember-a11y/core

PR #36, opened 2016-07-24
Fix parsing of custom violation class names

BrianSipple:
I discovered a fun bug with our custom `axeViolationClassNames` functionality:

We’re expecting this to be an Array when we use it in the `axe.a11yCheck` callback, but in the (I think very common) case where someone passes them as a string argument in a component’s template, that’s not quite what happens:

screen shot 2016-07-24 at 12 35 22 am
screen shot 2016-07-24 at 12 38 02 am

This fix for this is just to make sure we check for `axeViolationClassNames` being a string, and if it is, [split it on whitespace]() to make an array element from each word. 

I also took the opportunity to clean up the tests in general while writing a new one for this issue, so you’ll see a few line changes there as well. (The main change is a helper function that allows each test to not have to stub `axe.a11yCheck` manually.)

### DIFF
--- a/app/instance-initializers/axe-component.js
+++ b/app/instance-initializers/axe-component.js
@@ -40,12 +40,13 @@ export function initialize(application) {
     turnAuditOff: false,
 
     /**
-     * An array of classNames to add to the component when a violation occurs.
+     * An array of classNames (or a space-separated string) to add to the component when a violation occurs.
      * If unspecified, the `axe-violation` class is used to apply our default
      * styling
      *
      * @public
-     * @type {Array}
+     * @type {(Array|string)}
+     * @see(https://github.com/ember-a11y/ember-a11y-testing/blob/master/content-for/head-footer.html)
      */
     axeViolationClassNames: ['axe-violation'],
 
@@ -71,7 +72,12 @@ export function initialize(application) {
       if (this.get('tagName') !== '') {
         axe.a11yCheck(this.$(), this.axeOptions, (results) => {
           const violations = results.violations;
-          const violationClassNames = this.get('axeViolationClassNames');
+          let violationClassNames = this.get('axeViolationClassNames');
+
+          if (typeof violationClassNames === 'string') {
+            // support passing as a space-separated string
+            violationClassNames = violationClassNames.trim().split(/\s+/);
+          }
 
           for (let i = 0, l = violations.length; i < l; i++) {
             let violation = violations[i];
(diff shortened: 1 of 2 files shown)


BrianSipple:
@ember-a11y/core Friendly ping 😄.

binhums:
Sorry @BrianSipple, reviewing now

app/instance-initializers/axe-component.js
```diff
@@ -40,12 +40,13 @@ export function initialize(application) {
     turnAuditOff: false,
 
     /**
-     * An array of classNames to add to the component when a violation occurs.
+     * An array of classNames (or a space-separated string) to add to the component when a violation occurs.
```
binhums: Not unreasonable. If we're expecting people to use this as part of public API we should probably add an example to the docs.

binhums:
👍  I'd be happy to merge this, the code looks good.

I have some initial thoughts about the public API of the library. I actually think they're not super relevant because this won't have a big impact but I'd be interested in seeing your comments.

This is reminiscent of the `classNames` property on a vanilla component, not that I can speak for the motivation when this was initially written. That property [is restricted to arrays](https://github.com/emberjs/ember.js/blob/28cedad1418bedd2dad44eb56668fa67b9769d86/packages/ember-views/lib/mixins/class_names_support.js#L26).

If we're indeed looking for consistency with that API don't find myself wanting to specify `classNames` in handlebars templates, but specify it in the JS file for the component. When specifying classes in the template the `class` property is used (as per your example screenshot).

I guess my question is are the `axeViolationClassNames` likely to change by instance of the component in the template? Does it make sense to have something similar to the `class` property. I think that would be overkill, just initial musings.

BrianSipple:
Interesting questions.

I think this could make for a good exploration/discussion in its own issue thread, but I can certainly give my initial thoughts. 

I definitely _could_ see it being nice to have an API that behaves similarly. That said, we’d need to make sure that our concerns were analogous to those of the standard `Component` object. 

The way I see it now, `axeViolationClassNames` is kind of like `classNames`, but then I don’t know that we have the same additional concerns as `Component` -- in that `Component` seeks to enable a rich interface for designing a full UI system. And so I guess I see just having one property adapt to both use cases being enough atm, as opposed to having three different APIs a la `class`, `classNames`, and `classNameBindings`.

binhums:
:+1: Agreed :)

PR #38, opened 2016-08-07
Ensure axe-versions are in sync

sivakumar-kailasam:
Realized that the `axe-core` version in the blueprint is outdated. Fixed it & added a test to make sure we keep the version in the addon and blueprint in sync.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,3 +38,5 @@ script:
   # Usually, it's ok to finish the test scenario without reverting
   #  to the addon's original dependency state, skipping "cleanup".
   - ember try $EMBER_TRY_SCENARIO test --skip-cleanup
+  - npm install bower
+  - node node-tests/blueprint-tests.js
--- a/blueprints/ember-a11y-testing/index.js
+++ b/blueprints/ember-a11y-testing/index.js
@@ -4,6 +4,6 @@ module.exports = {
   normalizeEntityName: function() { },
 
   afterInstall: function() {
-    return this.addBowerPackageToProject('axe-core', '1.1.1');
+    return this.addBowerPackageToProject('axe-core', '~2.0.5');
   }
 };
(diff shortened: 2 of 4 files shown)


BrianSipple:
Nice catch -- and I love the idea of blueprint testing. But I'm wondering if we shouldn't separate our concerns here a bit more: one update for the instant fix, and another for a long-term testing strategy.

One reason being is that I think this could be a good opportunity to explore [ember-cli-blueprint-test-helpers](https://github.com/ember-cli/ember-cli-blueprint-test-helpers). This would allow us to leverage a standardized pattern for [building the addon](https://github.com/ember-cli/ember-cli-blueprint-test-helpers#embernewoptions), [generating the blueprint](https://github.com/ember-cli/ember-cli-blueprint-test-helpers#embergenerateargs), and then going to town with assertions. We then wouldn't need to have a special npm script in `package.json` and build step in `travis.yml`.

In any case, though, to at least address the version discrepancy quickly, it would be great to get @ember-a11y/core thoughts here on the best strategy going forward 😄.

binhums:
@sivakumar-kailasam Nice work!

@BrianSipple, do you want to raise a separate issue to track this?

sivakumar-kailasam:
@binhums @BrianSipple _ember-cli-blueprint-helpers_ atm doesn't support addons that use `addBowerPackageToProject`. I noticed it before writing these tests. Given the interest we have in it now, I've raised an [issue](https://github.com/ember-cli/ember-cli-internal-test-helpers/issues/22) with them. I'll get that into ember-cli-blueprint-test-helpers and replace these tests with the standard ones.

BrianSipple:
@sivakumar-kailasam Sweet. I'll open up another issue to track the replacement. 

With _all_ of that sorted out, I think we should merge what we have here and cut a new patch release. Replacing the tests later would just be another patch and we shouldn't have to worry about breaks, compatibility, etc. Thoughts @binhums?.

binhums:
Complete and utter agreement

PR #43, opened 2016-09-18
Separate route, template, and acceptance test for violations helper

BrianSipple:
A few wires got crossed during the last few merges, and tests for our violations helper in #35 weren't quite accounting for the deluge of components that had been added to our index page after #39. 

So, in addition to fixing a few JSHint errors, I created a separate `violations-helper` route/template so that we could have a more stable place to run an acceptance test for the helper. I also moved said tests into their own file.

This should get us back to green in CI.


MelSumner:
Closing this based on the amount of time that has passed without any action on it. Please feel free to re-submit the PR once all of the issues are resolved though! Thank you! :)

Issue #37, opened 2016-07-26
Add ember-cli-eslint


BrianSipple:
See: https://github.com/ember-a11y/issues/issues/2

binhums:
I'm happy to take this now if you're not working on it.

BrianSipple:
@binhums All yours!

BrianSipple:
@binhums ping 😛

trentmwillis:
Resolved by #66.

PR #48, opened 2016-12-02
Use ember-cli-eslint

billpull:
Seems like there are some tests failing being fixed in https://github.com/ember-a11y/ember-a11y-testing/pull/43 but this should be passing from an eslint perspective https://github.com/ember-a11y/ember-a11y-testing/issues/37


trentmwillis:
@billpull would you mind rebasing this? If not, I'll pick up this task.

trentmwillis:
I went ahead and finished this up in #66. Thanks for the initial work on this though! Greatly appreciated.

PR #53, opened 2017-02-13
Fix Travis and improve overall test stability

trentmwillis:
This should hopefully get CI green at which point we'll be able to land other PRs.

### DIFF
--- a/tests/integration/instance-initializers/axe-component-test.js
+++ b/tests/integration/instance-initializers/axe-component-test.js
@@ -1,9 +1,11 @@
 /* global sinon, axe */
 import Ember from 'ember';
 import { initialize } from 'dummy/instance-initializers/axe-component';
-import { module, test } from 'qunit';
+import { initialize as initializeViolationsHelper } from 'dummy/instance-initializers/violations-helper';
+import { moduleForComponent, test } from 'ember-qunit';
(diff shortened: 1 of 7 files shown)


tests/integration/instance-initializers/axe-component-test.js
```diff
@@ -146,120 +139,114 @@ test('audit should log any violations found', function(assert) {
...
 
 /* Component.axeCallback */
 
 test('axeCallback receives the results of the audit', function(assert) {
-  const results = { violations: [] };
-  const axeCallbackSpy = sandbox.spy();
-  const component = Component.create({
-    axeCallback: axeCallbackSpy
-  });
+  initialize();
```
rwjblue: Seems like this is always used, can we move it to the `beforeEach` instead of in each test?

rwjblue:
We should probably add ember-lts-2.8 to CI (basically supporting "last 2 LTS" in 2.4 + 2.8).

trentmwillis:
@ember-a11y/core would appreciate a review/approval from one of y'all before merging (though this doesn't change any actual behavior, just testing stuff)

trentmwillis:
Also got the :+1: from @MelSumner on Slack. Merging...

Issue #47, opened 2016-12-01
tests do not fail on a11y violations


sarbbottam:
the errors get logged into the browser's console (`ember test --serve`) but tests do not fail.

billpull:
@sarbbottam which version of ember-a11y-testing are you using? I was able to get my project to fail on errors by switching back to `0.1.4`

sarbbottam:
I'm using `0.1.6`.

billpull:
This seems to be an issue with `axe-core` for me anything above `1.1.1` simply logs the error to the console rather than reporting it to Qunit.

sarbbottam:
I tried with `axe-core@1.1.0` and the tests seems to fail. However, I can't seem to identify what is th reason for failure from the output.

```
ember test

Could not start watchman; falling back to NodeWatcher for file system events.
Visit http://ember-cli.com/user-guide/#watchman for more info.
Built project successfully. Stored in "/Users/sabandyo/GitCorp/ember-a11y-testing-demo/tmp/core_object-tests_dist-JsLmnkfl.tmp".
not ok 1 PhantomJS 2.1 - Global error: Error: Assertion Failed: The page should have no accessibility violations. Please check the developer console for more details. at http://localhost:7357/assets/vendor.js, line 17311
    ---
        Log: |
            { type: 'error',
              text: 'Error: Assertion Failed: The page should have no accessibility violations. Please check the developer console for more details. at http://localhost:7357/assets/vendor.js, line 17311\n' }
    ...
not ok 2 PhantomJS 2.1 - Acceptance | welcome: visiting /
    ---
        actual: >
            null
        stack: >
            http://localhost:7357/assets/vendor.js:17311
        message: >
            Error: Assertion Failed: The page should have no accessibility violations. Please check the developer console for more details.
        Log: |
    ...
ok 3 PhantomJS 2.1 - JSHint | acceptance/welcome-test.js: should pass jshint
ok 4 PhantomJS 2.1 - JSHint | app.js: should pass jshint
ok 5 PhantomJS 2.1 - JSHint | helpers/destroy-app.js: should pass jshint
ok 6 PhantomJS 2.1 - JSHint | helpers/module-for-acceptance.js: should pass jshint
ok 7 PhantomJS 2.1 - JSHint | helpers/resolver.js: should pass jshint
ok 8 PhantomJS 2.1 - JSHint | helpers/start-app.js: should pass jshint
[...]

PR #55, opened 2017-02-14
Add a11y-audit test helper

trentmwillis:
This introduces a test helper, `a11yAudit`, to run the axe-core audit on-demand within tests, instead of relying on the default auto-run or component-reopening.

cc @rwjblue @sarbbottam @ember-a11y/core

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,56 @@ You can see the available options in the [axe-core repo](https://github.com/dequ
 
 _Note:_ the options will stay set, until set to something different.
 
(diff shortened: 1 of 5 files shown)


trentmwillis:
@rwjblue only including you to verify my test helper setup is sane.

trentmwillis:
@ember-a11y/core would appreciate your opinion on whether this new API seems good. If it does, we can likely use the same helper in the auto-run scenario which should fix #47.

PR #57, opened 2017-02-14
Upgrade dependencies and improve stability of auto-run

trentmwillis:
Hoping this cuts down on the Travis flakiness.

Previously, the auto-run feature wasn't using a test waiter or test promises, and thus we had no guarantee that the audit actually finished prior to the test run finishing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,23 +9,23 @@ sudo: false
 
 cache:
   directories:
-    - node_modules
+    - $HOME/.npm
+    - $HOME/.cache # includes bowers cache
 
 matrix:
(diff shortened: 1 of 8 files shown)


trentmwillis:
@ember-a11y/core this appears to the prior flakiness I was seeing in Travis runs. A review would be appreciated before merging (you can make the diff much more manageable by appending `?w=1` to the url when viewing changes).

Robdel12:
the `?w=1` trick is ❤️

trentmwillis:
> the `?w=1` trick is ❤️

I know right?

Thanks for the review!

Issue #56, opened 2017-02-14
Remove axe-core as a Bower dependency


trentmwillis:
We should include axe-core via npm instead and add it via `treeForVendor`.

trentmwillis:
This has been fixed via #58.

PR #58, opened 2017-02-14
Move axe-core to npm dependency

trentmwillis:
Addresses #56.

This also allows us to remove our blueprint and related tests.


trentmwillis:
Rebased this on the other recent changes. Once again, a review from @ember-a11y/core folks would be appreciated 😄

PR #59, opened 2017-02-15
Properly report exceptions

trentmwillis:
Since axe-core switched to using Promises internally, the exceptions during auto-run were getting swallowed. There is no correct way to hook into the Testing Promise system Ember uses, so we instead simply catch any errors and report them through the test adapter.

This will fix https://github.com/ember-a11y/ember-a11y-testing/issues/47.

### DIFF
--- a/content-for/test-body-footer.html
+++ b/content-for/test-body-footer.html
@@ -19,8 +19,6 @@
        * @return {Void}
        */
       a11yCheckCallback: function(results) {
-        runningAudit = false;
-
         var violations = results.violations;
 
         if (violations.length) {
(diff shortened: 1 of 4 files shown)


content-for/test-body-footer.html
```diff
@@ -60,7 +58,10 @@
        */
       afterRender: function() {
         runningAudit = true;
-        axe.a11yCheck('#ember-testing-container', axe.ember.testOptions, axe.ember.a11yCheckCallback);
+        return axe.run('#ember-testing-container', axe.ember.testOptions)
+          .then(axe.ember.a11yCheckCallback)
+          .catch(Ember.Test.adapter.exception)
+          .then(() => runningAudit = false);
```
rwjblue: Should be a finally

trentmwillis: axe-core only uses ES6 Promises, which don't have `.finally` (yet, it's in proposal)

rwjblue: Gotcha! Thanks for double checking.

trentmwillis:
@ember-a11y/core, would appreciate another review. This should be the last one before releasing `v0.2.0`, which will hopefully make it much easier for folks to adopt (since it's more stable and has the on-demand `a11yAudit` helper).

trentmwillis:
Updated an existing test to verify the exception reporting behavior. Merging this now...

Issue #60, opened 2017-02-15
0.2.0: axe.run doesn't polyfill Promise implementation


trentmwillis:
Since axe.run doesn't polyfill a Promise implementation, if a browser doesn't support Promises natively and the consumer isn't providing a polyfill, things break. This wasn't caught in our test suite since we include the babel polyfill.

rwjblue:
"polyfill" it 😝:


```js
// vendor/ember-a11y-testing/promise-polyfill.js

(function() {
  if (typeof Promise === 'undefined') {
    self.Promise = Ember.RSVP.Promise;
  }
})();
```

Then use `app.import('vendor/ember-a11y-testing/promise-polyfill.js', { type: 'test' });` in `index.js`.

rwjblue:
The `app.import` above needs to go above the `app.import` for `vendor/axe-core/axe.js`.

trentmwillis:
@rwjblue I have tried something similar to the above, but the issue is that Ember.RSVP.Promise ties into the run-loop. Thus, since we trigger the audit "after render" in the run-loop, we actually wind up with an infinite loop (from what I can tell).

trentmwillis:
So in trying to fix this, I keep running upon other issues. In particular, changing the implementation from using a Promise to a using callbacks uncovers the fact that we run additional audits during the run-loop when the application is getting destroyed. This is bad, but we have no way to prevent it.

Due to that, the above issue, and other factors, I've opened an issue to remove the auto-run feature, which makes the path forward substantially simpler.

trentmwillis:
With the removal of auto-run in #64 and the usage of `RSVP.Promise` for the `a11yAudit` helper, this should be resolved.

PR #62, opened 2017-02-15
Include axe-core in all non-prod environments

### LINKED ISSUE #61: "ReferenceError: axe is not defined" when upgrading to 0.2.0
# Summary
When upgrading an app from ember-a11y-testing 0.1.6 to 0.2.0, I get a JS console error `ReferenceError: axe is not defined` when running `ember server`.

When I run unit tests (`ember test --server`), I do not get these errors and unit tests behave as expected. App uses Ember 2.8 (the newest LTS).

# Steps to reproduce / test case
See pull request demonstrating changes:
https://github.com/abought/demo-ember-osf/pull/6

# Attempted fixes
Also attempted to make package.json and bower.json changes to bring project in line with changes made here.
https://github.com/ember-a11y/ember-a11y-testing/pull/58

Got the same result both before and after `08238ff`.

Reran blueprint and tried with a fresh install of node_modules and bower_components.  

Tried in Chrome 56 and FF 51, Mac OS.

trentmwillis:
Fixes #61. Verified by linking locally against the reproduction repo.

This was introduced when moving `axe-core` to `npm` (#58).

cc @abought

### DIFF
--- a/index.js
+++ b/index.js
@@ -24,7 +24,9 @@ module.exports = {
   included: function(app) {
     this._super.included(app);
 
-    app.import('vendor/axe-core/axe.js', { type: 'test' });
+    if (app.env !== 'production') {
+      app.import('vendor/axe-core/axe.js');
+    }
   },
 
   treeForVendor: function() {


abought:
Thanks! 

I've confirmed this works for me in development, in that I no longer receive console errors about a missing module. (and that it warns me about accessibility violations in development mode)

trentmwillis:
Great! I'll be releasing a new version with this fix after fixing #60.

index.js
```diff
@@ -24,7 +24,9 @@ module.exports = {
   included: function(app) {
     this._super.included(app);
 
-    app.import('vendor/axe-core/axe.js', { type: 'test' });
+    if (app.env !== 'production') {
```
rwjblue: Should be if `app.tests` (which handles the `ember test --environment=production` case also)

Issue #17, opened 2016-05-19
Actually running in afterRender


nathanhammond:
Original issue: https://github.com/trentmwillis/ember-axe/issues/2

trentmwillis:
Auto-run has been removed so this issue is no longer valid.

PR #64, opened 2017-02-16
Remove auto run

### LINKED ISSUE #60: 0.2.0: axe.run doesn't polyfill Promise implementation

### LINKED ISSUE #63: Drop auto-run feature?
So, having spent the last couple days working on the addon and trying to fix issues like https://github.com/ember-a11y/ember-a11y-testing/issues/60, I am of the opinion that we should drop the "auto-run" feature for tests.

Here is my reasoning:

1. It makes adoption of the addon difficult because it will likely result in many failures for existing applications.
2. It can be quite spammy, since some applications trigger run-loops frequently in their app code, which amplifies point #1.
3. In addition to noise, running the audit on every turn of the run-loop can also be quite slow and costly for applications with complex UIs and many test cases.
4. The implementation of auto-run is hacky at best and has proven difficult to maintain, since we are effectively jamming outside, asynchronous behavior into the application's run-loop.
5. The implementation is currently QUnit-specific, which limits our ability to support those using other testing libraries.

Beyond that, I see several benefits to changing the approach to be on-demand (e.g., using `a11yAudit()`):

1. Has a clear, incremental path for adopting accessibility testing.
2. Is more explicit about when audits occur (less "magic" testing).
3. It is much easier to reason about the story for passing options and scoping the audit.
4. Will be much easier to maintain and test over time.

@ember-a11y/core thoughts? I'm especially curious if anyone is actually using the auto-run feature as it exists today. My hunch is that either no one is, or very few people are.

trentmwillis:
Overall points of this PR is to remove the auto-run feature as outlined in #63.

Other commits and reasoning:
- Remove babel polyfill from test app (this doesn't seem to be needed and was covering up issues like #60)
- Remove blueprint tests from npm scripts (these are no longer needed)
- Ensure axe-core is included in production test builds (just in case)

Closes #63, #17
Fixes #60


trentmwillis:
@ember-a11y/core reviews appreciated here. Once landed, I'd like to prioritize getting ESLint setup properly and once that lands we can cut a `0.3.0` release.

I think having linting in place and most of the current issues closed out will make it much easier for this project to continue iterating forward (which hopefully also means easier adoption from projects).

README.md
```diff
@@ -23,51 +21,12 @@ ember install ember-a11y-testing
...
-
-On the flip side, if you want to turn tests off, simply use:
-
-```javascript
-axe.ember.turnAxeOff();
-```
+Ember A11y Testing injects the `axe` global during development and tests so that
+you can run accessibility audits while developing your application. It also
+provides some Ember-specific tools so that you can integrate accessibility
+checks into your workflow easily.
```
binhums: May be worth providing example usage or a link to the API.

binhums:
Code itself looks great :+1:

I don't think I understand enough about this addon to comment on the features.

README.md
trentmwillis: We can update this in the future, but I feel like the examples further down should suffice.

trentmwillis:
@MelSumner and/or @Robdel12 review from one of you would be good before I merge

Issue #69, opened 2017-03-18
Integration test - "axeCallback throws an error if it is not a function" is failing


sarbbottam:
[axeCallback throws an error if it is not a function](https://github.com/ember-a11y/ember-a11y-testing/blob/master/tests/integration/instance-initializers/axe-component-test.js#L163-L170) is failing for me.

Anyone else experiencing the same?

```
not ok 37 PhantomJS 2.1 - Integration | Instance Initializer | axe-component: axeCallback throws an error if it is not a function
    ---
        actual: >
            false
        expected: >
            true
        stack: >
            exception@http://localhost:7357/assets/test-support.js:14246:49
            adapterDispatch@http://localhost:7357/assets/vendor.js:50285:22
            dispatchError@http://localhost:7357/assets/vendor.js:28554:23
            invokeWithOnError@http://localhost:7357/assets/vendor.js:10918:14
            flush@http://localhost:7357/assets/vendor.js:10974:15
            flush@http://localhost:7357/assets/vendor.js:11098:20
            end@http://localhost:7357/assets/vendor.js:11168:28
            run@http://localhost:7357/assets/vendor.js:11282:19
            run@http://localhost:7357/assets/vendor.js:33259:32
            render@http://localhost:7357/assets/test-support.js:15143:30
            http://localhost:7357/assets/tests.js:575:19
            throws@http://localhost:7357/assets/test-support.js:4613:17
            http://localhost:7357/assets/tests.js:574:18
            runTest@http://localhost:7357/assets/test-support.js:3700:34
            run@http://localhost:7357/assets/test-support.js:3686:13
            http://localhost:7357/assets/test-support.js:3864:15
            process@http://localhost:7357/assets/test-support.js:5098:26
            begin@http://localhost:7357/assets/test-support.js:5081:11
            http://localhost:7357/assets/test-support.js:4298:11
        message: >
            Error: Assertion Failed: axeCallback should be a function.
        Log: |
            { type: 'error', text: 'null\n' }
```

<img width="1058" alt="screen shot 2017-03-17 at 7 37 23 pm" src="https://cloud.githubusercontent.com/assets/949380/24068323/39e00238-0b49-11e7-82fa-58e03bc21834.png">

**Steps followed**
1. forked and cloned the fork
2. `npm i && bower install`
3. `ember test`

trentmwillis:
[...]

PR #70, opened 2017-03-20
feat(a11yAudit): opt in to a11yAudit

sarbbottam:
Added `a11yAuditIf` method to run `a11yAudit` if `enableA11yAudit=true` is passed as `--query`.

---

**Context**

We want to setup a11y audit as a nightly offline process and don't want to disrupt the current CI/CD pipeline. Since the code changes to accommodate the a11y audit is done in the same repository, we wanted to have a controlled method.

So for the regular acceptance test run `a11yAuditIf()` will function as a `noop` since `--query=enableA11yAudit=true` will not be available.

However for, the dedicated nightly build due to the presence of `--query=enableA11yAudit=true` as part of the test command a11y audit will be performed.

### DIFF
--- a/addon-test-support/audit-if.js
+++ b/addon-test-support/audit-if.js
@@ -1,0 +1,30 @@
+import RSVP from 'rsvp';
+import a11yAudit from './audit';
+import utils from './utils';
+
+/**
+ * A method to return the value of queryParameter
+ *
+ * @method getUrlParameter
+ * @private
+ */
+function getUrlParameter(name) {
+    const location = utils.getLocation();
+    const regex = new RegExp('[\\?&]' + name + '=([^&#]*)');
+    const results = regex.exec(location.search);
+    return results === null ? '' : decodeURIComponent(results[1].replace(/\+/g, ' '));
+};
+
+/**
+ * A wrapper method to run the a11yAudit if desired
+ *
+ * @method a11yAuditIf
+ * @public
+ */
+export default function a11yAuditIf(...args) {
+  if(getUrlParameter('enableA11yAudit') === 'true') {
+    return a11yAudit(...args);
+  }
+
+  return RSVP.resolve(undefined, 'a11y audit not run');
+}
--- a/addon-test-support/utils.js
+++ b/addon-test-support/utils.js
@@ -1,0 +1,5 @@
+export default {
+  getLocation() {
+    return window && window.location;
+  }
+};
--- a/tests/integration/helpers/a11y-audit-if-test.js
+++ b/tests/integration/helpers/a11y-audit-if-test.js
@@ -1,0 +1,37 @@
+import Ember from 'ember';
+import { moduleForComponent, test } from 'ember-qunit';
+import hbs from 'htmlbars-inline-precompile';
+import utils from 'ember-a11y-testing/test-support/utils';
+import a11yAuditIf from 'ember-a11y-testing/test-support/audit-if';
+
+const { Component } = Ember;
+
+// We use a component integration test to verify the behavior of the a11y-audit
+// by rendering a component and then running the audit on it.
+moduleForComponent('component:axe-component', 'Integration | Helper | a11y-audit', {
+  integration: true,
+
+  beforeEach() {
+    this.register('component:axe-component', Component.extend());
+  }
+});
+
+test('a11yAuditIf should not execute a11yAudit', function(assert) {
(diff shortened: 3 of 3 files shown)


addon-test-support/audit.js
```diff
@@ -66,10 +66,41 @@ Ember.Test.registerAsyncHelper('a11yAudit', function(app, ...args) {
...
 }
+
+/**
+ * A wrapper method to run the a11yAudit if desired
+ *
+ * @method a11yAuditIf
+ * @public
+ */
+function a11yAuditIf(...args) {
+  if(getUrlParameter('enableA11yAudit') === 'true') {
```
sarbbottam: Wanted to cache the output of `getUrlParameter('enableA11yAudit')` however phantom keeps complaining 
```
Error: Browser disconnected
```

tests/acceptance/a11y-audit-test.js
```diff
@@ -1,7 +1,7 @@
 import Ember from 'ember';
 import { module, test } from 'qunit';
 import startApp from '../../tests/helpers/start-app';
-import a11yAudit from 'ember-a11y-testing/test-support/audit';
+import { a11yAudit } from 'ember-a11y-testing/test-support/audit';
```
sarbbottam: I could not figure out how to simulate the availability of the `enableA11yAudit` param, thus have not written any test for the newly added `a11yAuditIf`. Any pointer might help me to write tests for `a11yAuditIf`.

addon-test-support/audit.js
```diff
@@ -66,10 +66,41 @@ Ember.Test.registerAsyncHelper('a11yAudit', function(app, ...args) {
...
+ * @private
+ */
+function getUrlParameter(name) {
+    name = name.replace(/[\[]/, '\\[').replace(/[\]]/, '\\]');
+    var regex = new RegExp('[\\?&]' + name + '=([^&#]*)');
+    var results = regex.exec(location.search);
+    return results === null ? '' : decodeURIComponent(results[1].replace(/\+/g, ' '));
+};
+
+export {
```
trentmwillis: Changing these exports will be a breaking change, meaning we would need to release a new major version, which I would rather not do. Instead, we should extract the new function into it's own module.

tests/acceptance/a11y-audit-test.js
trentmwillis: I would write a helper to retrieve `window.location` which you can then stub. Like so:

```js
export {
  getLocation() {
    return window && window.location;
  }
};
```
Then in the test:
```js
import helper from '../helper'; // come up with better names here
helper.getLocation = function() { retun mockLocationObject; }
```

trentmwillis:
I've also fixed #69, so if you rebase you should be back to "green".

sarbbottam:
@trentmwillis I updated as per review comments, could you take another look?

addon-test-support/audit-if-utils.js
```diff
@@ -0,0 +1,5 @@
+export default {
```
trentmwillis: Can we rename this file to just `utils`?

addon-test-support/audit-if.js
```diff
@@ -0,0 +1,31 @@
...
+import utils from './audit-if-utils';
+
+/**
+ * A method to return the value of queryParameter
+ *
+ * @method getUrlParameter
+ * @private
+ */
+function getUrlParameter(name) {
+    name = name.replace(/[\[]/, '\\[').replace(/[\]]/, '\\]');
```
trentmwillis: Can we remove this? Since we know that this function isn't going to be used for all kinds of `name`s, I think this is unnecessary.

addon-test-support/audit-if.js
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * @method a11yAuditIf
+ * @public
+ */
+export default function a11yAuditIf(...args) {
+  if(getUrlParameter('enableA11yAudit') === 'true') {
+    return a11yAudit(...args);
+  }
+
+  return RSVP.resolve('no operation');
```
trentmwillis: You can set a label for the Promise, which seems to be what you really want here:

```js
return RSVP.resolve(undefined, 'a11y audit not run');
```

sarbbottam:
Updated 😄

sarbbottam:
Thanks @trentmwillis

Issue #77, opened 2017-04-07
ember-a11y-testing make a few things scroll at the bottom of the page.


AdrienNguyenWorldline:
If I leave the plugin active, when I click on some actions, I get no error in console at all, nor stiped elements, but the page scrolls at the bottom.

If I just add `turnAuditOff: true,` then this scrolling bugs goes away

trentmwillis:
Can you provide a reproduction either via ember-twiddle or a github repo? This addon doesn't do anything directly related to scrolling.

justwright:
I think this may be related to scrolling issues within axe-core

https://github.com/dequelabs/axe-core/pull/148

ctcpip:
Not sure if the exact same issue, but I observe this as well.  It is automatically scrolling to the last ember component on the page.

ctcpip:
The axe-core issue mentioned above has been fixed & merged and does not resolve this issue.

Another situation that easily reproduces the issue (and it's even worse):  I am displaying a relative time - example: '3 seconds ago', '10 seconds ago', etc.  This value is updated every second, and every time it is updated, `ember-a11y-testing` scrolls to its position.  It even happens when the text doesn't even change - for example, it simply shows '3 minutes ago' for an entire minute, but nonetheless `ember-a11y-testing` scrolls to this line every second. Very frustrating.  `turnAuditOff` does prevent the behavior, but that somewhat defeats the purpose of using this addon.

ctcpip:
```
Ember      : 2.13.0
Ember Data : 2.13.1
jQuery     : 3.2.1
```

dylanb:
to fix this issue, you need to pass in the `noScroll` option to the `color-contrast` rule

ctcpip:
@dylanb Thanks for that tip!  Can confirm that it works.  It took me a while to figure out the right syntax... the `axe-core` docs were not very helpful.  Here it is:

```
  ENV['ember-a11y-testing'] = {
    componentOptions: {
      axeOptions: {
        checks: {
          'color-contrast': { options: { noScroll: true } }
        }
      }
    }
  };
```

ctcpip:
@trentmwillis I suggest this option be set by default within this addon if possible...  this made me pull my hair out, and I think others will have the same experience...  I don't think this scrolling is a desirable default behavior.  Thoughts?

trentmwillis:
Currently, we leave the defaults to `axe-core`, so if this is the default for that library I'd prefer not to change it.

[...]

PR #82, opened 2017-06-22
README - example to disable color contrast scroll

ctcpip:
update README.md - include example code for disabling the scroll option for the color contrast check - (#77)

### DIFF
--- a/README.md
+++ b/README.md
@@ -271,6 +271,20 @@ ENV['ember-a11y-testing']: {
 }
 ```
 
+The color contrast check can cause undesired scrolling behavior.  To disable:
+
+```javascript
+ENV['ember-a11y-testing'] = {
+  componentOptions: {
+    axeOptions: {
+      checks: {
+        'color-contrast': { options: { noScroll: true } }
+      }
+    }
+  }
+};
+```
+
 ## Future Plans
 
 Now that your components and acceptance tests can self-audit, the next step
--- a/README.md
+++ b/README.md
@@ -257,32 +257,21 @@ ENV['ember-a11y-testing'] = {
 };
 ```
 
-By example, to disable a specific check you can:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -271,6 +271,20 @@ ENV['ember-a11y-testing']: {
 }
 ```
 
+The color contrast check can cause undesired scrolling behavior.  To disable:
```
trentmwillis: I'd prefer to avoid having two examples right next to each other that are essentially the same thing. Can we modify the example above instead? Maybe we can word it as "For example, the `color-contrast` rule has some default scrolling functionality which you may want to disable like so:"

ctcpip: The first example disables a rule and the other specifies an option for a check.  (As an aside, I'm still unclear on what the difference is between a check and a rule.)  In any case, I'm fine with however you want to proceed.  I submitted the PR because of how much hair-pulling was involved to get this figured out per #77, so I think it's justified to have examples in the readme.  Another idea could be to combine the examples (so there's only one `ENV['ember-a11y-testing'] = {` and possibly add comments for each directive.  Of course, then we'd have to change the rule that is disabled since we are specifying an option on the rule (check?) we just disabled...  Thoughts?

trentmwillis: We can combine the examples (which was what I meant by my above suggestion, I guess I was just unclear).

The purpose of this document should be to show how to use the features provided by ember-a11y-testing, which, in this case, amounts to passing options through to the underlying axe-core library. We should provide an example of how exactly to specify the options and then defer to the [axe-core documentation for specifics](https://github.com/dequelabs/axe-core/blob/develop/doc/API.md#options-parameter).

ctcpip: Agreed.  I'll take a stab at it.

trentmwillis:
Looks good to me, thanks!

PR #84, opened 2017-10-05
Package upgrades and locking version of htmlbars-inline-precompile.

renatoi:
- Enforcing installation of axe-core 2.4.2 or greater due to security issues.
- Locked ember-cli-htmlbars-inline-precompile to 0.3.11. Versions greater than 0.3.11 (0.3.12 and 0.3.13) are causing template compilation to fail (hence all pull requests since last updates were failing).

@trentmwillis


trentmwillis:
Released in version `0.4.4`

PR #89, opened 2017-11-28
Allow nested addon.

camskene:
If an app consumes another addon that has ember-a11y-testing as a `dependency`  ember-a11y-testing will be available in that app.

### DIFF
--- a/index.js
+++ b/index.js
@@ -42,7 +42,7 @@ module.exports = {
    * @override
    */
   contentFor: function(type) {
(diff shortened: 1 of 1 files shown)


camskene:
My use case was I have several apps that have a common dependency - rather than adding ember-a11y-testing to every app I can add it once to that dependency.

I noticed a drawback after submitting this PR; in development the background 'violation' styles don't come through. I will close this PR and re-open until if/when I fix that. 

Works fine if you just want to use it in testing.

PR #94, opened 2018-02-16
Update Ember-CLI

Robdel12:
## What is this?

This PR updates Ember-CLI to 2.15.x to help solve some of the CI issues we're seeing on PRs. This also sets the addon up for an easier upgrade path to 3.x when we cross that bridge.


tests/dummy/app/index.html
```diff
@@ -9,8 +9,8 @@
 
     {{content-for "head"}}
 
-    <link rel="stylesheet" href="{{rootURL}}assets/vendor.css">
-    <link rel="stylesheet" href="{{rootURL}}assets/dummy.css">
+    <link integrity="" rel="stylesheet" href="{{rootURL}}assets/vendor.css">
```
MelSumner: what does this do?

tests/dummy/config/targets.js
```diff
@@ -0,0 +1,9 @@
+/* eslint-env node */
+module.exports = {
+  browsers: [
+    'ie 9',
```
MelSumner: is it still IE9?

Robdel12: Yea, I think we merge this & do a major version bump if we want to change which browsers we support (which is perfect timing with Ember 3.x :D)

tests/dummy/app/index.html
Robdel12: It's supposed to let the browser check to see if the code was manipulated. (I didn't know before so I googled):

> Integrity attribute is to allow the browser to check the file source to ensure that the code is never loaded if the source has been manipulated.

https://stackoverflow.com/a/32042283/1352387

PR #96, opened 2018-03-28
add async/await syntax to readme (closes #95)

### LINKED ISSUE #95: Update to new async syntax
Examples in the docs like

```javascript
import a11yAudit from 'ember-a11y-testing/test-support/audit';

// ...elided for brevity

test('Some test case', function(assert) {
  visit('/');
  a11yAudit();
  andThen(() => assert.ok(true, 'no a11y errors found!'));
});
```
should be updated to:
```javascript
import a11yAudit from 'ember-a11y-testing/test-support/audit';

// ...elided for brevity

test('Some test case', async function(assert) {
  visit('/');
  await a11yAudit();
  assert.ok(true, 'no a11y errors found!');
});
```
...to follow the new async syntax. And maybe add a note how to use this addon for anyone still using the 2x branch. If this change is okay, I can create a PR with the changes

atrama:
Added async/await syntax to README.md and kept Ember 2x andThen() syntax for users still on 2x version.


README.md
```diff
@@ -39,6 +39,20 @@ import a11yAudit from 'ember-a11y-testing/test-support/audit';
...
 // ...elided for brevity
 
+test('Some test case', async function(assert) {
+  visit('/');
+  await a11yAudit();
+  assert.ok(true, 'no a11y errors found!'));
+});
+```
+
+If you are using Ember 2.18 or below, you will need to use the the andThen() helper and a regular function (as opposed to `async function`).
```
rwjblue: Hmm, you can use `await` going back many versions. It shouldn’t be tied to 2.18.

trentmwillis: Also, shouldn't `visit()` use `await`?

atrama: Thanks, I'll try and update both of these this week. @rwjblue I believe the only pre-req to using this syntax is using `ember-cli-qunit` 4.2? I think that became default in `ember-cli` 2.18 but anyone using older version can update manually. Do I have that correct?

rwjblue: ember-cli-qunit@4.2 allowed using the newer testing styles proposed in [emberjs/rfcs#268](https://github.com/emberjs/rfcs/blob/master/text/0268-acceptance-testing-refactor.md) and [emberjs/rfcs#232](https://github.com/emberjs/rfcs/blob/master/text/0232-simplify-qunit-testing-api.md), however you could still use `await visit(...)` using the (now legacy) `moduleForAcceptance` API's as long as your application allows async/await in general.

atrama: Items updated. Sorry for the delay. A new baby since the previous commit severely cut into free time

README.md
```diff
@@ -60,9 +74,9 @@ test('Some test case', function(assert) {
     }
   };
 
-  visit('/');
-  a11yAudit(axeOptions);
-  andThen(() => assert.ok(true, 'no a11y errors found!'));
+  await visit('/');
+  await a11yAudit(axeOptions);
+  assert.ok(true, 'no a11y errors found!'));
```
jenweber: There's one too many parens here: `assert.ok(true, 'no a11y errors found!'));`

Congrats on the new baby!!!

atrama: Oops, I’ll get that fixed this week or weekend

atrama:
Is there anything blocking this PR from being merged? All checks above have passed.

PR #103, opened 2018-05-17
Improve logging of violations in test mode

EndangeredMassa:
This PR extracts the violation formatting logic into a util class that can be used in both component audit and test audit scenarios.

It also changes the violationsHelper so that it works more like a singleton and is used throughout the project as a module instead of a global. It's still exposed as a global in the initializer, though. This also makes it work properly in test mode.

---

Here's a before/after of using ember-a11y-testing in another project's test suite:

Before:

After:

---

Here's a screenshot of development mode to show that it still works fine:

### DIFF
--- a/addon-test-support/audit.js
+++ b/addon-test-support/audit.js
@@ -3,6 +3,8 @@ import { assert } from '@ember/debug';
 import Ember from 'ember';
 import RSVP from 'rsvp';
 import config from 'ember-get-config';
+import formatViolation from 'ember-a11y-testing/utils/format-violation';
+import violationsHelper from 'ember-a11y-testing/utils/violations-helper';
 
 /**
  * Processes the results of calling axe.a11yCheck. If there are any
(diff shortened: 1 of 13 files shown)


dbollinger:
Looks good-- you are a saint! Any reason not to continue outputting the total number of violations?

EndangeredMassa:
Added a violation count to the error message. It now looks like:

```
[serious]: Elements must have sufficient color contrast 
Violated 5 times. Offending nodes are: 
<span class="c-radio-track__item-node">None</span>
<span class="c-radio-track__item-node">Level 1</span>
<span class="c-radio-track__item-node">Level 2</span>
<span class="c-radio-track__item-node">Level 3</span>
<p data-test-selector="labeless-text-input" id="violations__low-contrast-text" class="p-violations__grid-item-content--low-contrast-text c-paragraph ember-view">          Swoooosh
</p>
https://dequeuniversity.com/rules/axe/2.6/color-contrast?application=axeAPI
{object}
```

I also changed `Offending markup is` to `Offending nodes are` because what follows is a list of every node that violated that rule.

JoaoGFarias:
Thank you! This improvement makes checking and debugging a11y in Ember apps so much faster!

Issue #102, opened 2018-05-14
visitWithAudit helper


ondrejsevcik:
Hi, it would be nice to have a helper where the `visit` and `a11yAudit` are combined into one so I don't forget to run audit. 

```
// before
await visit('/');
await a11yAudit();

// after
await visitWithAudit('/');
```

Maybe there is a better name than `visitWithAudit`, but you got the idea ;)

rwjblue:
Going to close this one (sorry @ondrejsevcik), I think we should refocus and target some specific features to do this via the new (private for now) hooks added in emberjs/ember-test-helpers#878.

PR #105, opened 2018-05-17
Add a helper for visit-and-audit

EndangeredMassa:
This is a request for feedback on a new helper. After the details are decided, I'll update the readme to reflect its usage.

This PR adds a new helper called `visitAndAudit`. It simply visits a page, then audits it. This allows you to more easily always run an audit in acceptance tests, for example.

I tried using this in an app that uses the newer `import { visit } from '@ember/test-helpers';` to get a reference to visit, which works. I'm not sure if this is the best way to support multiple ways to get `visit` or not.

---

This works in a consuming app by:

```
import visit from 'ember-a11y-testing/test-support/visit-and-audit';
```

Then using `visit` (or whatever you want to call it) like normal:

```
visit('/products');
```

Or with aXe options:

```
visit('/products', axeOptions);
```

---

Resolves: https://github.com/ember-a11y/ember-a11y-testing/issues/102

### DIFF
--- a/addon-test-support/visit-and-audit.js
+++ b/addon-test-support/visit-and-audit.js
@@ -1,0 +1,15 @@
+import { visit as visitHelper } from '@ember/test-helpers';
+import a11yAudit from './audit';
+
+/**
+ * A wrapper method to run the a11yAudit if desired
+ *
+ * @method a11yAuditIf
+ * @public
+ */
+export default function visitAndAudit(path, axeOptions) {
+  let visit = window.visit || visitHelper;
+  return visit(path).then(() => {
+    return a11yAudit(axeOptions);
+  });
+}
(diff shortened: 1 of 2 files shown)


addon-test-support/visit-and-audit.js
```diff
@@ -0,0 +1,15 @@
...
+import a11yAudit from './audit';
+
+/**
+ * A wrapper method to run the a11yAudit if desired
+ *
+ * @method a11yAuditIf
+ * @public
+ */
+export default function visitAndAudit(path, axeOptions) {
+  let visit = window.visit || visitHelper;
```
EndangeredMassa: Is there a better way to support the two testing methods?

EndangeredMassa:
I'll do some more manual testing linking this to different kinds of projects before merging

MelSumner:
@EndangeredMassa are you okay with this being merged?

EndangeredMassa:
@MelSumner no. I want to test it some more. I'm not sure it works properly in some older test setups.

rwjblue:
Going to close this one (sorry @EndangeredMassa), I think we should refocus and target some specific features to do this via the new (private for now) hooks added in https://github.com/emberjs/ember-test-helpers/pull/878.

PR #109, opened 2018-09-24
Axe-core v3.1.2 update

drewlee:
There have been quite a number of axe-core updates since v2.6.1, including numerous bug fixes, additional rules and rule updates, and performance optimizations. As such, it is imperative that an organization getting started with accessibility testing is able to take advantage of new features provided in the latest release of axe-core.

This pull request facilitates the axe-core version upgrade to 3.1.2. Axe-component.js has been updated to use the newer axe.run API method (a11yCheck has been deprecated), and updated the callback function signature (first argument is now an error object, second argument is the audit result). Related tests have also be updated to reflect these changes.

One particular aspect to take note related to axe-core v3, is that concurrent axe calls, such as those that are made in the component audit, result in numerous exceptions as detailed under https://github.com/dequelabs/axe-core/issues/1041. I've created a new service (concurrent-axe.js) which mitigates the issue by deferring subsequent axe.run calls. Fortunately, this issue does not occur when axe is run from within tests.

I've also fixed the demo page, which didn't seem to properly update the UI with its corresponding visual noise level. The issue was occurring due to the fact that the violations-grid-item component was also executing an axe audit on its content, which in turn override the nested component's visuals. This was easily mitigated by disabling audit on the violations-grid-item component. Additionally, the "Level 3" checkbox was broken due to incorrect ID assignment.

I've verified that all changes pass tests and demo page works as expected.

No Visual Noise
Visual Noise Level 1
Visual Noise Level 2
Visual Noise Level 3

### DIFF
--- a/addon/services/concurrent-axe.js
+++ b/addon/services/concurrent-axe.js
@@ -1,0 +1,41 @@
+/* global axe */
+import Service from '@ember/service';
+import { next } from '@ember/runloop';
+
+export default Service.extend({
+  init() {
+    this._super(...arguments);
+    this._timer = null;
+    this._queue = [];
+  },
+
+  /**
+   * Axe v3 contains a concurrency issue which breaks the component auditing feature.
+   * This service defers axe.run calls on onto the next loop so that concurrent 
+   * axe executions do not occur.
+   *
+   * @see(https://github.com/dequelabs/axe-core/issues/1041)
+   * @public
+   * @param {HTMLElement} element axe context
+   * @param {Object} options axe configuration options
+   * @param {Function} callback axe audit callback
+   * @return {Void}
+   */
+  run(element, options, callback) {
+    if (this._timer) {
+      this._queue.push(arguments);
+    } else {
+      this._timer = next(() => {
+        if (element && element.parentNode) {
+          axe.run(element, options, callback);
+        }
+
+        this._timer = null;
+
+        if (this._queue.length) {
+          this.run(...this._queue.shift());
+        }
+      });
+    }
+  }
+});
--- a/app/instance-initializers/axe-component.js
+++ b/app/instance-initializers/axe-component.js
@@ -9,6 +9,7 @@ import ENV from '../config/environment';
 import isBackgroundReplacedElement from 'ember-a11y-testing/utils/is-background-replaced-element';
 import formatViolation from 'ember-a11y-testing/utils/format-violation';
 import violationsHelper from 'ember-a11y-testing/utils/violations-helper';
+import { inject as service } from '@ember/service';
 
 const VIOLATION_CLASS__LEVEL_1 = 'axe-violation--level-1';
 const VIOLATION_CLASS__LEVEL_2 = 'axe-violation--level-2';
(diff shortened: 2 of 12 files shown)


drewlee:
Looks like the Travis CI build is failing due to a transient dependency. execa@1.0.0 expects a Node version of at least 6. The Node dependency version should probably be bumped up to 6, as ember-cli also now expects v6 as the minimum.

addon/services/concurrent-axe.js
```diff
@@ -0,0 +1,38 @@
...
+   * @param {HTMLElement} element axe context
+   * @param {Object} options axe configuration options
+   * @param {Function} callback axe audit callback
+   * @return {Void}
+   */
+  run(element, options, callback) {
+    if (this._timer) {
+      this._queue.push(arguments);
+    } else {
+      this._timer = next(() => {
```
rwjblue: I'm slightly concerned about the async added here. For example:

* What happens when the element is removed from the DOM before the audit has occurred? Does axe care if the element that is audited is in-dom? 
* How do we ensure the element state is consistent (between when the audit was started and when it is ran)?

app/instance-initializers/axe-component.js
```diff
@@ -54,6 +55,15 @@ export function initialize() {
...
 
   Component.reopen({
+    /**
+     * An Ember service which resolves concurrent axe.run v3 issue.
+     *
+     * @public
+     * @type {Object}
+     * @see(https://github.com/dequelabs/axe-core/issues/1041)
+     */
+    concurrentAxe: service(),
```
rwjblue: Can you tweak this to:

```js
concurrentAxe: service('concurrent-axe'),
```

I know that with the default resolver the argument to `inject.service()` is optional here, but there are a few alternate resolvers out there that are much more strict (and would not normalize `concurrent-axe` into `concurrentAxe`).

app/instance-initializers/axe-component.js
```diff
@@ -138,12 +148,16 @@ export function initialize() {
...
       if (this.get('tagName') !== '') {
-
-        axe.a11yCheck(this.$(), this.axeOptions, (results) => {
+        this.get('concurrentAxe').run(this.element, this.axeOptions, (error, results) => {
           if (this.get('isDestroyed')) {
             return;
           }
 
+          if (error) {
+            Ember.Logger.error(error);
```
rwjblue: Based on the axe-core API docs, we expect `error` to be uncommon (and if it isn't we are instructed to report issues to axe-core itself). Since this is an exceptional case, can you update this to `throw error` instead of logging (and then swallowing) the error?

app/instance-initializers/axe-component.js
drewlee: Of course, should be fixed now.

app/instance-initializers/axe-component.js
drewlee: Yes, makes sense. Should be fixed now.

addon/services/concurrent-axe.js
drewlee: Removal of the component root element can definitely be problematic. As such, I've added a check to ensure that the element is still in the DOM before it is passed off to aXe (and also updated corresponding tests). Unfortunately, ensuring consistent element state is problematic even without the async, as the state can change between when aXe initially starts the audit on a component and when it completes. This concern is somewhat negligible given the nature of the tool - as it's well understood it will not catch 100% of all accessibility violations. A few milliseconds between operations would likely not incur drastic state changes, especially given that this is for developer use (I'd hope other developers wouldn't be interacting with a component before its corresponding audit completed). Additionally, the audit would rerun on each subsequent re-render.

rwjblue: Ok, seems good enough to me 😃. We can always circle back and attempt to address if there is any actual issue here.

drewlee: Yes, thanks Robert!

drewlee:
Anyone else (especially project maintainers) have feedback (or can facilitate the merge request)? If not, I will do so. I'll leave this up for another couple days, but will go ahead with the merge if I don't hear otherwise. Thanks!

PR #110, opened 2018-10-17
Fix services for engines, add option to exclude axe-core for visual component audit feature

drewlee:
Using a [service](https://github.com/ember-a11y/ember-a11y-testing/blob/master/addon/services/concurrent-axe.js) to mitigate the [aXe concurrency issue](https://github.com/ember-a11y/ember-a11y-testing/blob/master/addon/services/concurrent-axe.js) proved to be problematic for apps utilizing Ember engines, as the service needs to be explicitly defined as a dependency for each engine. This a bit too much overhead, especially for apps that don't utilize the visual component audit feature. As a fix, I have changed the service to a simple class, as there's no stateful information being stored in the service anyway. This is a much cleaner approach and removes the overhead of explicit dependency declarations.

We've also encountered a use case where we need to exclude axe-core from development builds (only required during testing). Similarly, for those not readily utilizing the visual component audit feature, including axe-core and modifying the Component class is a bit too much overhead. As such, I've added a new option for excluding axe-core if the visual component audit feature isn't being utilized.

### DIFF
--- a/addon/utils/concurrent-axe.js
+++ b/addon/utils/concurrent-axe.js
@@ -1,13 +1,11 @@
 /* global axe */
-import Service from '@ember/service';
 import { next } from '@ember/runloop';
 
-export default Service.extend({
-  init() {
-    this._super(...arguments);
+export class ConcurrentAxe {
+  constructor() {
     this._timer = null;
     this._queue = [];
-  },
+  }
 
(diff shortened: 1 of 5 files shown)


drewlee:
Thanks Robert!

PR #114, opened 2018-11-05
Fixes #113

### LINKED ISSUE #113: axe.run arguments are invalid on development use
Hi, thanks for this awesome addon.

I'm currently using `ember-a11-testing` **0.5.4**, everything works fine within test environment, but when I try to enable the audit within `development` mode I got this error :point_down: 

image

My current configuration is

```js
/* config/environment.js */

module.exports = function (environment) {
  const ENV = {
  .......
    'ember-a11y-testing': {
      componentOptions: {
        turnAuditOff: false, // Change to false to enable the audit in development
      }
    },


  };

  if (environment === 'development') {
    loggers(ENV);
  }

  .....
  return ENV;
};
```

This was working over a month ago when I added this addon, I'm not sure if #109 broke it in some aspect? :thinking: 

Thanks

drewlee:
Provides empty object if axeOptions are not defined to prevent axe.run from failing. Prior versions of axe worked with undefined options argument, while v3 explicitly expects an object. Resolves the issue by adding empty object default.

### DIFF
--- a/app/instance-initializers/axe-component.js
+++ b/app/instance-initializers/axe-component.js
@@ -45,7 +45,7 @@ export function initialize() {
   const {
     turnAuditOff: configuredTurnAuditOff,
     excludeAxeCore,
-    axeOptions,
+    axeOptions = {},
     axeCallback,
     visualNoiseLevel: configuredVisualNoiseLevel,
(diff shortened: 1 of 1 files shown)


Robdel12:
You're the best @drewlee ❤️

PR #115, opened 2018-11-19
Adds user timing mechanism for performance profiling

drewlee:
We'd like to add some simple user timing markers to get a better idea of a11y testing perf impacts on acceptance test suites.

### DIFF
--- a/addon-test-support/audit.js
+++ b/addon-test-support/audit.js
@@ -5,6 +5,7 @@ import RSVP from 'rsvp';
 import config from 'ember-get-config';
 import formatViolation from 'ember-a11y-testing/utils/format-violation';
 import violationsHelper from 'ember-a11y-testing/utils/violations-helper';
+import { mark, markEndAndMeasure } from './utils';
 
 /**
  * Processes the results of calling axe.a11yCheck. If there are any
@@ -53,6 +54,8 @@ function isPlainObj(obj) {
  * @private
  */
 function runA11yAudit(contextSelector = '#ember-testing-container', axeOptions) {
+  mark('a11y_audit_start');
+
   // Support passing axeOptions as a single argument
   if (arguments.length === 1 && isPlainObj(contextSelector)) {
     axeOptions = contextSelector;
@@ -80,7 +83,10 @@ function runA11yAudit(contextSelector = '#ember-testing-container', axeOptions) 
 
   return auditPromise
     .then(a11yAuditCallback)
-    .finally(() => document.body.classList.remove('axe-running'));
+    .finally(() => {
+      document.body.classList.remove('axe-running');
+      markEndAndMeasure('a11y_audit', 'a11y_audit_start', 'a11y_audit_end');
+    });
 }
 
 // Register an async helper to use in acceptance tests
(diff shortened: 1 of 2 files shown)


addon-test-support/audit.js
```diff
@@ -80,7 +87,26 @@ function runA11yAudit(contextSelector = '#ember-testing-container', axeOptions)
...
+ * @param {String} comment
+ * @param {String} startMark
+ * @param {String} endMark
+ * @return {Void}
+ * @private
+ */
+function markPerf(comment, startMark, endMark) {
+  if (performance) {
+    performance.mark(endMark);
+    measure(comment, startMark, endMark);
```
rwjblue: I believe this needs to be guarded in a `try` / `catch`. The rough `tldr;` is that anyone at any time can call `performance.clearMarks()` on us, and then calling `performance.measure(...)` here would throw an error.

Some additional context in https://github.com/qunitjs/qunit/pull/1319.

drewlee: Yup, this is being done here: https://github.com/ember-a11y/ember-a11y-testing/pull/115/files#diff-140c735f8ae8e94da2968b4bb2f97b7dR48, where the measure fn is defined 🙂

addon-test-support/utils.js
```diff
@@ -1,5 +1,54 @@
...
+   * @return {Void}
+   * @public
+   */
+  measure(comment, startMark, endMark) {
+    // `performance.measure` may fail if the mark could not be found.
+    // reasons a specific mark could not be found include outside code invoking `performance.clearMarks()`
+    try {
+      performance.measure(comment, startMark, endMark);
+    } catch (e) {
+      Ember.Logger.warn('performance.measure could not be executed because of ', e.message);
```
rwjblue: Ember.Logger is deprecated (I think the deprecation was added in 3.1 or 3.2), we should swap to `console.warn`.

See [the deprecation guide](https://emberjs.com/deprecations/v3.x#toc_ember-console-deprecate-logger) for more info.

addon-test-support/audit.js
rwjblue: Shoot. I'm sorry for not spotting that before, I thought I did search for where `measure` was coming from 😞

addon-test-support/audit.js
```diff
@@ -5,6 +5,9 @@ import RSVP from 'rsvp';
 import config from 'ember-get-config';
 import formatViolation from 'ember-a11y-testing/utils/format-violation';
 import violationsHelper from 'ember-a11y-testing/utils/violations-helper';
+import utils from './utils';
+
+const { performance, measure } = utils;
```
rwjblue: Maybe we could import these as named functions (instead of destructuring), what do you think?

```js
import { mark, measure } from './performance-utils';
```

We could even make those methods resilient to when `performance` is not defined, so we could remove the conditionals in this file...

addon-test-support/audit.js
drewlee: No worries.

addon-test-support/utils.js
drewlee: Cool, will do!

drewlee: Fixed.

addon-test-support/audit.js
drewlee: Yes, that's much cleaner now.

PR #118, opened 2019-03-08
Axe-core version bump to 3.2.1

drewlee:
Bumping axe-core to latest release, which has improved performance to speed up a11y audits. Also including a comment for visual audit instance initializer based on discussions with @stefanpenner.


drewlee:
Deque just released 3.2.2, so updated to match.

PR #130, opened 2019-08-21
Fixes #107

### LINKED ISSUE #107: Uncaught Error: Could not find module `@ember/debug` imported from `appName/instance-initializers/axe-component`
Installing ember-a11y-testing@0.5.2 on Ember app with ember-cli@2.12 stops page rendering and gives hard error: Uncaught Error: Could not find module `@ember/debug` imported from `appName/instance-initializers/axe-component`

Installing an older version such as 0.5.0 works fine.

drewlee:
Moves axe-component.js and violations-helper.js from app directory to addon directory. Replaces with re-exports to reflect the conventional structure of an ember addon.


drewlee:
One thing to note, looks like it's not possible to import the consuming app's config from the addon directory, so I've had to remove this import:
`import ENV from '../config/environment';`

and use this methodology instead:
`const ENV = appInstance.resolveRegistration('config:environment');`
 
This was based on the discussion at https://discuss.emberjs.com/t/best-practices-accessing-app-config-from-addon-code/7006. Seems to work well, but folks smarter than me are welcome to pipe up if this is problematic.

rwjblue:
Thanks for working on this @drewlee!

PR #132, opened 2019-09-04
Fixes #124. Allows a11yAudit to account for axe include/exclude context param.

### LINKED ISSUE #124: a11yAudit fails to account for include/exclude context parameter
In addition to a query selector or HTML DOM node, the axe context argument can also include an "include/exclude" pojo. The a11yAudit helper fails to account for this argument format and assumes that the object is the options argument.

For example, the following invocation will execute on the entire document instead of the assigned scope, and will fail to apply the axe configurations set in environment.js:
```
await a11yAuditIf({
  include: [['.col-thread']],
  exclude: [
    ['[data-control-name="select_gif"]']
  ]
});
```

The current workaround is to assign `null` as the second argument to bypass the internal formatting check:
```
await a11yAuditIf({
  include: [['.col-thread']],
  exclude: [
    ['[data-control-name="select_gif"]']
  ]
}, null);
```

drewlee:
Adds fix by checking for include and/or exclude properties. Adds test to verify correct behavior.

### DIFF
--- a/addon-test-support/audit.js
+++ b/addon-test-support/audit.js
@@ -45,6 +45,18 @@ function isPlainObj(obj) {
 }
 
 /**
+ * Determines whether supplied object contains `include` and `exclude` axe
+ * context selector properties. This is necessary to distinguish an axe
+ * config object from a context selector object, after a single argument
+ * is supplied to `runA11yAudit`.
+ * @param {Object} obj
+ * @return {Boolean}
+ */
+function isNotSelectorObj(obj) {
+  return !obj.hasOwnProperty('include') && !obj.hasOwnProperty('exclude');
+}
+
+/**
  * Runs the axe a11y audit with the given context selector and options.
  * The context defaults to '#ember-testing-container' if not specified.
  * The options default axe-core defaults.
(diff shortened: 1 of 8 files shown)


addon-test-support/audit.js
```diff
@@ -44,6 +44,18 @@ function isPlainObj(obj) {
...
+/**
+ * Determines whether supplied object contains `include` and `exclude` axe
+ * context selector properties. This is necessary to distinguish an axe
+ * config object from a context selector object, after a single argument
+ * is supplied to `runA11yAudit`.
+ * @param {Object} obj
+ * @return {Boolean}
+ */
+function isNotSelectorObj(obj) {
+  return !obj.hasOwnProperty('include') && !obj.hasOwnProperty('exclude');
```
Robdel12: Just want to double check that both `include` and `exclude` will always be present, right?

drewlee: We can use either/or. The possible combinations can be just `include`, just `exclude`, both `include` and `exclude`, or neither one. So this particular conditional will return true if neither is present, but false if either one or both are detected. I have tested all the permutations to make sure this behaves as expected. Let me know if you have additional questions. Thanks!

Robdel12: That works for me! Great stuff 😃

Issue #108, opened 2018-09-14
code is included in production builds


ctcpip:
Is it possible to exclude this code from being included when running `ember b --environment=production` ?

rwjblue:
Hmm, that should already be the default behavior...

ctcpip:
_"should"_ being the key word there :)

![screen shot 2018-09-14 at 3 03 13 pm](https://user-images.githubusercontent.com/15326526/45572609-58f19080-b82f-11e8-9df2-454d0e11849f.png)

rwjblue:
D'oh! Looks like we need [this kind of thing](https://github.com/ember-a11y/ember-a11y-testing/blob/750c2e316c4feed66b182ded3bdd6f1752246388/index.js#L54-L66) for `treeForAddon`...

ctcpip:
```
"addon-tree-output/ember-a11y-testing/utils/is-background-replaced-element.js":   1264,
--
"addon-tree-output/ember-a11y-testing/utils/violations-helper.js":   2892, |  
```

drewlee:
I have a fix for this in PR #134

drewlee:
Fixed w/PR #134. Will publish a new release sometime before Monday.

PR #134, opened 2019-09-27
Fixes code included in production builds [#108]

drewlee:
Adds override to exclude all addon related code from production builds.

Makes axe check a bit more defensive.

- We recently ran into an issue where the axe-component initializer was getting bundled with a bunch of engines, causing exceptions by calling axe where it didn't exist. Although this was ultimately a build issue not related to the addon, the failure can definitely be made much more graceful.

### DIFF
--- a/index.js
+++ b/index.js
@@ -59,11 +59,34 @@ module.exports = {
   treeForApp: function(tree) {
     var checker = new VersionChecker(this);
     var isProductionBuild = process.env.EMBER_ENV === 'production';
-    var isOldEmber = checker.for('ember', 'bower').lt('1.13.0');
+    var isOldEmber = checker.forEmber().lt('1.13.0');
 
     if (isProductionBuild || isOldEmber) {
       tree = new Funnel(tree, {
-        exclude: [/instance-initializers\/axe-component|violations-helper.js/]
+        exclude: [/instance-initializers\/(axe-component|violations-helper)\.js/]
+      });
+    }
+
+    return tree;
+  },
+
+  /**
+   * Exclude all addon code during build if this is a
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -63,7 +63,30 @@ module.exports = {
...
+  /**
+   * Exclude all addon code during build if this is a
+   * production build or if the version of Ember being used is less than 1.13.
+   * @override
+   */
+  treeForAddon: function() {
+    var tree = this._super.treeForAddon.apply(this, arguments);
+    var checker = new VersionChecker(this);
+    var isProductionBuild = process.env.EMBER_ENV === 'production';
+    var isOldEmber = checker.for('ember', 'bower').lt('1.13.0');
```
rwjblue: There is a special check added in ember-cli-version-checker for Ember specifically, so you don't have to specify bower vs npm:

```suggestion
    var isOldEmber = checker.forEmber().lt('1.13.0');
```

index.js
```diff
@@ -63,7 +63,30 @@ module.exports = {
...
+
+    if (isProductionBuild || isOldEmber) {
+      tree = new Funnel(tree, {
```
rwjblue: Why isn't this tree returned? I think just building it but not returning it will have no effect...

index.js
drewlee: Thanks for the suggestion @rwjblue. I'll get this updated.

index.js
drewlee: It is being returned on line 93. It's just the way the diff is being displayed, makes it not very apparent.

rwjblue: D'oh! Sorry about that!

PR #187, opened 2020-08-20
Attempting to fix renovate so it won't open pin dependency PRs

scalvert:
Here we go again...

I believe this will fix our issue 🤞

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,6 +1,7 @@
 {
   "extends": [
     "config:base",
+    ":preserveSemverRanges",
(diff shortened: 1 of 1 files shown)


scalvert:
cc/ @rwjblue

renovate.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": [
     "config:base",
+    ":preserveSemverRanges",
```
rwjblue, line 4: This is a wild preset name...

PR #199, opened 2020-08-22
Refactors param normalization in `a11yAudit`

scalvert:
This PR prepares for the potential of deprecating the `a11yAuditIf` helper. This helper can be merged into `a11yAudit` by allowing it to accept another optional parameter: URL. This allows for two things:

1. The aforementioned deprecation of an extra, largely unnecessary API (simplification)
2. The removal of the `getLocation` function, which returns `window.location` and exists only to provide a stubbing mechanism for the `getUrlParameter` function. By adding an extra URL param to `a11yAudit`, we can simply test that function by passing in the required value, and provide the default as `window.location`.

⚠️ This change also includes a breaking change that removes support for older test infra versions (folks using the long deprecated `moduleForAcceptance` test system) by removing `window.a11yAudit` support.

### DIFF
--- a/addon-test-support/audit.ts
+++ b/addon-test-support/audit.ts
@@ -14,7 +14,7 @@ type MaybeContextObject = ElementContext | RunOptions | undefined;
  * @param {Object} results
  * @return {Void}
  */
-function a11yAuditCallback(results: AxeResults) {
+function processAxeResults(results: AxeResults) {
   let violations = results.violations;
 
   if (violations.length) {
@@ -114,7 +114,7 @@ export default function a11yAudit(
     });
   });
 
-  return auditPromise.then(a11yAuditCallback).finally(() => {
+  return auditPromise.then(processAxeResults).finally(() => {
     document.body.classList.remove('axe-running');
     markEndAndMeasure('a11y_audit', 'a11y_audit_start', 'a11y_audit_end');
   });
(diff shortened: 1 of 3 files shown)


addon-test-support/audit.ts
```diff
@@ -71,67 +124,28 @@ function isNotContextObject(obj: MaybeContextObject) {
...
- * testing situation, but also supports being used in integration/unit test
- * scenarios.
- *
- * @method a11yAudit
- * @public
- */
-export default function a11yAudit(...args: any[]) {
-  if ((<any>window).a11yAudit) {
-    return (<any>window).a11yAudit(...args);
-  }
```
rwjblue, line 134: Should probably call this change out in the PR description, this is a "breaking" change (drops support for super old versions of the test harness that were relying on `Ember.Test.registerHelper`). We should totally clean it up (like you did here), just need to call it out explicitly.

addon-test-support/audit.ts
```diff
@@ -71,67 +124,28 @@ function isNotContextObject(obj: MaybeContextObject) {
...
-
-  return runA11yAudit(...args);
+  })
+    .then(processAxeResults)
```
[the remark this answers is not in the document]
scalvert: I agree. I have some ideas how to improve this. Let's for sure follow up.

addon-test-support/audit.ts
scalvert: Done.

Issue #186, opened 2020-08-20
[New API] Add `setupGlobalA11yHooks` helper


scalvert:
# Summary

This issue proposes adding a new global setup helper: `setupGlobalA11yHooks`. This helper will allow you to, with a single call in your test-helper.js file (or wherever makes sense), wire up the `a11yAudit(If)` calls to specific `@ember/test-helpers` hooks using the new [`_registerHooks` API](https://github.com/emberjs/ember-test-helpers/pull/878).

# Motivation

In order to add accessibility-specific checks in your tests, you need to explicitly import and call one of the `a11yAudit(If)` helpers. While the explicitness provides value - you can clearly see these audits are being called within specific tests - it's explicitness is also a drawback. Requiring engineers to remember and intentionally call these APIs can mean that these audits may fall by the wayside and be forgotten.

In order to encourage, and indeed elevate, accessibility audits as part of what is considered standard testing best practices, this PR proposes adding this additional mechanism to invoke these calls. This will allow tests to take advantage of these audits automatically.

# Details

As previously mentioned, this issue proposes adding the following:

**A new global helper exported from this addon called `setupGlobalA11yHooks`**

```ts
function setupGlobalA11yHooks(invocationStrategy: () => boolean): void;
``` 

- `invocationStrategy` - a `function` that returns a `boolean`, which controls _how_ the audit is invoked. This would control whether the audit occurs or not
	- should provide default `invocationStrategy` implementation:
		- `invokeAll`: invokes all audits
- should gracefully handle the case where the version of `@ember/test-helpers` does not contain the `_registerHook` API
- Call `_registerHook` from `@ember/test-helpers` to hook into test helpers to run `a11yAudit(If)` checks at specific times (after those helpers' invocations have completed by passing the 'end' hook as the second parameter to `_registerHook`)
- Register hooks (initially) for
	- `visit`
	- `click`
	- `doubleClick`
	- `tap`
- Register non-`a11yAudit` hooks for (helpers that don't need to run `a11yAudit` calls, but can benefit from additional validations, such as whether outer label elements are present, etc.
	- `fillIn`
	- `typeIn`
- Provide a mechanism to provide configuration to the individual `a11yAudit(If)` calls. Since engineers won't be invoking these functions themselves, we still need a way to pass options into these calls at the individual test level. We can provide a `setA11yAuditOptions` function that takes these options, and is cleaned up at the end of each test run.
[...]

PR #200, opened 2020-08-25
Implements `setupGlobalA11yHooks` helper

scalvert:
Implementation of #186 

Does not include (should be done in a separate PR):
- Register non-a11yAudit hooks for (helpers that don't need to run a11yAudit calls, but can benefit from additional validations, such as whether outer label elements are present, etc.
	- `fillIn`
	- `typeIn`


addon-test-support/audit-if.ts
```diff
@@ -27,10 +27,10 @@ function getUrlParameter(name: string) {
 export default function a11yAuditIf(
   contextSelector?: ElementContext | RunOptions | undefined,
   axeOptions?: RunOptions | undefined
-) {
+): Promise<void> {
```
rwjblue, line 30: FWIW, I'd rather use PromiseLike :P

addon-test-support/invocation-strategies.ts
```diff
@@ -0,0 +1,21 @@
+export interface InvocationStrategy {
+  (): boolean;
+}
+
+export const invokeAll: InvocationStrategy = function invokeAll(): boolean {
+  return true;
+};
+
+export function invokeEveryN(interval: number): InvocationStrategy {
```
rwjblue, line 15: Maybe `invokePercentange`?

tests/dummy/app/index.html
```diff
@@ -16,6 +16,12 @@
   </head>
   <body>
     {{content-for "body"}}
+    <div id="qunit"></div>
+    <div id="qunit-fixture">
+      <div id="ember-testing-container">
+        <div id="ember-testing"></div>
+      </div>
+    </div>
```
rwjblue: Should be removed, this should only be present in tests/index.html

addon-test-support/invocation-strategies.ts
scalvert: I felt like every n tests was easier to think about. Why do you prefer percentage?

tests/dummy/app/index.html
scalvert: This was a mistake, and I already removed it.

addon-test-support/audit-if.ts
scalvert: Yes, we've had this conversation before. I can probably do that, as in this case we don't need any other functions in our promise usage.

addon-test-support/invocation-strategies.ts
scalvert: In fact, I think `n tests` is more precise, as it's actually invoking every n test. No fake news :trollface:!

addon-test-support/run-options.ts
```diff
@@ -0,0 +1,32 @@
...
+ */
+export function getRunOptions(context: object) {
+  return _runOptions.get(context);
+}
+
+/**
+ * Sets up the cleanup of the run options for a specific test.
+ */
+export function setupRunOptions() {
+  QUnit.testDone(function () {
```
rwjblue, line 29: I don't really want to bake in `QUnit` semantics here

addon-test-support/run-options.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+
+/**
+ * Gets run options specific to a test.
+ *
+ * @param context Test context object, accessed using `@ember/test-helpers` `getContext` function.
+ */
+export function getRunOptions(context: object) {
+  return _runOptions.get(context);
+}
```
rwjblue, line 23: I think I'd prefer to use a stack like system here. What do you think?

addon-test-support/run-options.ts
scalvert: Ya I'm with you. I'm writing this while we're speaking about using the destroyables APIs. I agree. :)

addon-test-support/setup-global-a11y-hooks.ts
```diff
@@ -0,0 +1,40 @@
...
+ * @param audit Optional audit function used to run the audit. Allows for providing either a11yAudit, a11yAuditIf,
+ *              or custom audit implementation.
+ */
+export function setupGlobalA11yHooks(
+  shouldAudit: InvocationStrategy,
+  audit: (...args: any[]) => PromiseLike<void> = a11yAudit
+) {
+  ['visit', 'click'].forEach((helperName) => {
+    let hook = _registerHook(helperName, 'end', async () => {
+      if (shouldAudit()) {
```
rwjblue, line 23: I think shouldAudit should receive the helper name / label, so you could implement a custom per-helper system...

addon-test-support/setup-global-a11y-hooks.ts
```diff
@@ -0,0 +1,40 @@
...
+        await audit(getRunOptions(getContext()));
```
rwjblue, line 24: I'd like to remove these arguments, and call `audit(helperName, label)` here too

addon-test-support/setup-global-a11y-hooks.ts
scalvert: I like. Adding.

addon-test-support/setup-global-a11y-hooks.ts
scalvert: I think we need these for now. Let's explore removing this when we remove the `ember-get-config` (as we discussed).

addon-test-support/run-options.ts
scalvert: I like this. This makes perfect sense to me in that the top of the stack will contain the specific options we're using. So, the algo will look like the following:

calling `setRunOptions`...

- in `tests/test-helper.js` pushes default options onto the stack. If nothing else pushes options, the defaults are used
- in test module's `beforeEach`, pushes specific options onto the stack. If nothing else pushes options, these options are used. These options need to be popped off the stack once the module is done
- test pushes specific options onto the stack. Uses these options, as they're the top of the stack. These options need to be popped off the stack once the test is done.

addon-test-support/invocation-strategies.ts
scalvert: We've moved this to an example instead.

addon-test-support/run-options.ts
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Sets run options specific to a test.
+ *
+ * @param options Axe {RunOptions} to be provided to the audit helper.
+ */
+export function setRunOptions(options: RunOptions = {}) {
+  optionsStack.push(options);
+
+  registerDestructor((getContext() as any).owner, () => optionsStack.pop());
```
scalvert, line 15: I'm not 100% that this is correct. I'm not sure that attaching the destructor to the owner is enough to correctly attribute the options to all levels of options (global, test module level, test level). Would the global invocation even have an owner?

tests/acceptance/setup-global-a11y-hooks-test.ts
```diff
@@ -0,0 +1,109 @@
...
+
+    hooks.beforeEach(function () {
+      setupGlobalA11yHooks(invokeAll, a11yAuditFake);
+    });
+
+    hooks.afterEach(function () {
+      teardownGlobalA11yHooks();
+    });
+
+    Array(TEST_INVOCATIONS_COUNT)
```
drewlee: We should probably add a comment or two about why this is being done this way.

addon-test-support/setup-global-a11y-hooks.ts
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * @param shouldAudit Invocation strategy function that determines whether to run the audit helper or not.
+ * @param audit Optional audit function used to run the audit. Allows for providing either a11yAudit, a11yAuditIf,
+ *              or custom audit implementation.
+ */
+export function setupGlobalA11yHooks(
+  shouldAudit: InvocationStrategy,
+  audit: (...args: any[]) => PromiseLike<void> = a11yAudit
+) {
+  ['visit', 'click'].forEach((helperName) => {
```
drewlee, line 19: This is just a general thought and not related to the code at hand. Would the host app be able to unregister a hook, if they choose to do so? I'm thinking that perhaps a particular test does some setup work after a `visit`, making the `a11yAudit` not particularly useful at that point in time.

addon-test-support/index.ts
```diff
@@ -0,0 +1,8 @@
+export { default as a11yAudit } from './audit';
+export { default as a11yAuditIf } from './audit-if';
+export { invokeAll } from './invocation-strategies';
+export { setRunOptions, getRunOptions } from './run-options';
+export {
+  setupGlobalA11yHooks,
+  teardownGlobalA11yHooks,
+} from './setup-global-a11y-hooks';
```
drewlee, line 8: Just to better my knowledge, what is the reason for having to re-export these modules? Thanks!

tests/acceptance/setup-global-a11y-hooks-test.ts
scalvert: What do you mean specifically? Comments don't age well, so I'd prefer to use comments sparingly.

What do you think will help here that the code can't tell you?

addon-test-support/setup-global-a11y-hooks.ts
scalvert: Can you think of a specific use case we would need to support? I'd be helpful to have a specific use case to work from so we can determine the best API.

addon-test-support/index.ts
scalvert: We re-export in the index of the addon (or in this case, the addon's `addon-test-support` dir), so that you can reference those imports using a single, public path. This means you won't have to 'dig in' to the module to find specific things to import, but rather the index file indicates what's publicly available. It also makes import paths a bit cleaner. Does this make sense?

drewlee: Yes, totally. Thanks for explaining!

tests/acceptance/setup-global-a11y-hooks-test.ts
drewlee: Defining a test from within an iterator is an atypical pattern. I'm thinking that it would be helpful to provide some context, in that this is being done to verify invoking an nth test.

addon-test-support/setup-global-a11y-hooks.ts
drewlee: Modals and similar types of overlays are common examples, which require a set of user actions to render on screen. We would essentially end up with some overlap if a prior test covers the main page. However, I don't think this is something we should tackle at this point, as we're going to run into cases of overlap, regardless.

tests/acceptance/setup-global-a11y-hooks-test.ts
scalvert: I'm not sure it's an atypical pattern. We do this in [other places](https://github.com/ember-lifeline/ember-lifeline/blob/master/tests/unit/dom-event-listeners-test.js#L46) to test variants. In this case, we're testing the test infrastructure in order to validate that our helpers work (this addon is a test infra addon). I don't believe it warrants a comment, but could understand expanding the test name if you have suggestions.

addon-test-support/setup-global-a11y-hooks.ts
scalvert: What are the drawbacks of having overlaps if that results in less overhead for the engineer writing the tests? Is it a performance concern with reading DOM?

drewlee: One is performance, the other is mitigating overly noisy output when it comes to a11y violations. However, the latter can likely be addressed via deduping violation output.

addon-test-support/invocation-strategies.ts
```diff
@@ -0,0 +1,10 @@
+export interface InvocationStrategy {
+  (helperName: string, label: string): boolean;
+}
+
+/**
+ * Invocation strategy implementation to run audits all tests.
+ */
+export const invokeAll: InvocationStrategy = function invokeAll(): boolean {
+  return true;
+};
```
rwjblue, line 10: Can you remove this file?

addon-test-support/run-options.ts
```diff
@@ -0,0 +1,27 @@
...
+ * Sets run options specific to a test.
+ *
+ * @param options Axe {RunOptions} to be provided to the audit helper.
+ */
+export function setRunOptions(options: RunOptions = {}) {
+  optionsStack.push(options);
+
+  if (getContext()) {
+    registerDestructor((getContext() as any).owner, () => optionsStack.pop());
+  }
```
rwjblue, line 17:
```suggestion
  let context = getContext();
  if (context) {
    registerDestructor((context as any).owner, () => optionsStack.pop());
  }
```

addon-test-support/should-force-audit.ts
```diff
@@ -0,0 +1,13 @@
+let _url: URL = new URL(window.location.href, document.baseURI);
```
rwjblue, line 1:
```suggestion
const _url: URL = new URL(window.location.href, document.baseURI);
```

package.json
```diff
@@ -65,7 +67,7 @@
     "ember-export-application-global": "^2.0.1",
     "ember-load-initializers": "^2.1.1",
     "ember-maybe-import-regenerator": "^0.1.6",
-    "ember-qunit": "^4.6.0",
+    "ember-qunit": "5.0.0-beta.2",
```
rwjblue, line 70: Should probably be `^` 

```suggestion
    "ember-qunit": "^5.0.0-beta.2",
```

tests/acceptance/violations-test.js
```diff
@@ -15,5 +18,7 @@ module('Acceptance | temp test', function (hooks) {
       violationsHelper.count > 0,
       'Violations are found in the violationsHelper'
     );
+
+    setEnableA11yAudit(false);
```
rwjblue, line 22: Should either be in a `finally` or in `hooks.afterEach`

tests/integration/helpers/a11y-audit-if-test.js
```diff
@@ -35,5 +32,7 @@ module('Integration | Helper | a11yAuditIf', function (hooks) {
       );
       assert.ok(foundExpectedError);
     }
+
+    setEnableA11yAudit(false);
```
rwjblue, line 36: should be in finally or afterEach

addon-test-support/invocation-strategies.ts
scalvert: Yep. Done.

addon-test-support/setup-global-a11y-hooks.ts
```diff
@@ -0,0 +1,38 @@
...
+ * @param audit Optional audit function used to run the audit. Allows for providing either a11yAudit, a11yAuditIf,
+ *              or custom audit implementation.
+ */
+export function setupGlobalA11yHooks(
+  shouldAudit: InvocationStrategy,
+  audit: (...args: any[]) => PromiseLike<void> = a11yAudit
+) {
+  ['visit', 'click', 'doubleClick', 'tap'].forEach((helperName) => {
+    let hook = _registerHook(helperName, 'end', async () => {
+      if (shouldForceAudit() && shouldAudit(helperName, 'end')) {
```
drewlee, line 21: Curious as to the reason for the `shouldForceAudit()` check? As far as I can tell, this determines whether the `enableA11yAudit` query param is set. However, since the consumer can specify which variant to run (`a11yAudit`, `a11yAuditIf`, or custom method), shouldn't it be only relevant for the `a11yAuditIf` invocation? For `a11yAuditIf`, the check would effectively occur twice. For all other cases, we would need to either set the query param in the URL, or run `setEnableA11yAudit(true)` in the hooks.

scalvert: We want to deprecate `a11yAuditIf`. We can achieve that functionality in a different way, specifically though this new util.

drewlee: Yes, I get that. I just don't think `shouldForceAudit` should be a requisite for running `setupGlobalA11yHooks`, especially when using a custom audit implementation. Ultimately though, I'll leave this up to your discretion. Thanks!

rwjblue: Without something like `shouldForceAudit` it is impossible for the addon to instruct folks how to reproduce a failure if they had used a `shouldAudit` that "randomizes". Having something that randomly fails tests, but is difficult/impossible to reproduce is **massively frustrating**. This `shouldForceAudit` does exactly that, it ensures that if you _want_ to run A11y audits, that you can reliably do it.

addon-test-support/should-force-audit.ts
```diff
@@ -0,0 +1,13 @@
...
+  if (enabled) {
+    _url.searchParams.set('enableA11yAudit', 'true');
+  } else {
+    _url.searchParams.delete('enableA11yAudit');
+  }
+}
+
+export function shouldForceAudit() {
+  return _url.searchParams.get('enableA11yAudit') !== null;
+}
```
rwjblue, line 13: I think this is static, and won't work. The `_url` variable won't live update when the running URL bar changes...

```suggestion
export function shouldForceAudit() {
  const url = new URL(window.location.href, document.baseURI);

  return url.searchParams.get('enableA11yAudit') !== null;
}
```

addon-test-support/should-force-audit.ts
```diff
@@ -0,0 +1,13 @@
+const _url: URL = new URL(window.location.href, document.baseURI);
+
+export function setEnableA11yAudit(enabled: boolean = false) {
+  if (enabled) {
+    _url.searchParams.set('enableA11yAudit', 'true');
+  } else {
+    _url.searchParams.delete('enableA11yAudit');
+  }
+}
```
rwjblue, line 9: I think this has conceptually the same issue that I mentioned in the `shouldForceAudit` section. It happens to work because you share the same (stale) `_url` object, but that won't work in general.

```suggestion
export function setEnableA11yAudit(enabled: boolean = false) {
  const url = new URL(window.location.href, document.baseURI);

  // set up the enableA11yAudit query param
  if (enabled) {
    url.searchParams.set('enableA11yAudit', '');
  } else {
    url.searchParams.delete('enableA11yAudit');
  }
  
  // updates the URL without reloading
  window.history.replaceState(null, '', url.href)
}
```

tests/integration/helpers/a11y-audit-if-test.js
```diff
@@ -16,24 +18,23 @@ module('Integration | Helper | a11yAuditIf', function (hooks) {
...
-      assert.ok(foundExpectedError);
+      try {
+        await a11yAuditIf(this.element);
+        assert.ok(false, 'audit should have failed');
+      } catch (error) {
+        let foundExpectedError = error.message.startsWith(
+          'The page should have no accessibility violations.'
+        );
+        assert.ok(foundExpectedError);
+      }
```
rwjblue, line 35: Can also be moved to `assert.rejects` (seems totally fine for a future PR)

addon-test-support/setup-global-a11y-hooks.ts
drewlee: Yeah, I guess I'm just thrown off with `?enableA11yAudit=true` being used for more than one purpose. Either way, you guys have way more insight into the testing infra space, so it's not something I'm gonna push.

tests/integration/helpers/a11y-audit-if-test.js
scalvert: Fixed it for completeness.

addon-test-support/setup-global-a11y-hooks.ts
rwjblue: Ya, I definitely see what you mean there. I think that @scalvert and I are trying to move to a space where there is only one meaning (force running the audits). This includes a bunch of steps on the path though:

* stop suggesting actual test code use `a11yAudit` or `a11yAuditIf`
* remove the a11yAuditIf helper
* this implementation (for the hooks system)
* some additional implementation for customized validation of interaction helpers (fillIn, typeIn, &co) - likely a future PR
* move the non-testing code out of the addon (we totally need a nice dev mode experience, but it belongs in a separate addon)

Once all that stuff is done, I don't think we have that consistency problem where the query param means two things...

PR #201, opened 2020-08-26
Converts the `Ember.assert` usage to throwing error for failing tests

scalvert:
This addon was leveraging Ember's `assert` function to trigger failures. This PR removes that in favor of plain `Error` objects (which is ultimately what `assert` uses under the hood.

### DIFF
--- a/tests/acceptance/a11y-audit-test.js
+++ b/tests/acceptance/a11y-audit-test.js
@@ -16,15 +16,11 @@ module('Acceptance | a11y audit', function (hooks) {
 
     await visit('/');
 
-    try {
-      await a11yAudit();
-      assert.ok(false, 'a11yAudit should have thrown an error on violations');
-    } catch (error) {
-      let foundExpectedError = error.message.startsWith(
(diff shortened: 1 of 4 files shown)


tests/acceptance/a11y-audit-test.js
```diff
@@ -21,7 +21,7 @@ module('Acceptance | a11y audit', function (hooks) {
       assert.ok(false, 'a11yAudit should have thrown an error on violations');
     } catch (error) {
       let foundExpectedError = error.message.startsWith(
-        'Assertion Failed: The page should have no accessibility violations. Violations:'
+        'The page should have no accessibility violations. Violations:'
       );
       assert.ok(foundExpectedError, 'error message is correct');
     }
```
rwjblue, line 27: Can you refactor this to use `assert.rejects`?

scalvert: Oh absolutely. I thought of using that, but just didn't go down that path. Will do!

PR #203, opened 2020-08-28
Deprecate `a11yAuditIf`

scalvert:
Adds a deprecation warning for `a11yAuditIf` and integrates that functionality into `a11yAudit`. The latter will become the primary primitive to invoke audits.


addon-test-support/audit-if.ts
```diff
@@ -13,6 +14,15 @@ export default function a11yAuditIf(
...
   axeOptions?: RunOptions | undefined
 ): PromiseLike<void> {
+  deprecate(
+    "ember-a11y-testing's `a11yAuditIf` is deprecated. Please use a11yAudit when using the `enableA11yAudit` query param",
+    false,
+    {
+      id: 'ember-a11y-testing-deprecated-a11y-audit-if',
+      until: '5.0.0',
+    }
+  );
```
rwjblue, line 24: Maybe we could tell folks to themselves use `forceA11yAudit() && await a11yAudit();` themselves? I mean, its kinda silly, but it should work...

addon-test-support/audit.ts
```diff
@@ -131,6 +132,13 @@ export default function a11yAudit(
 ): PromiseLike<void> {
   mark('a11y_audit_start');
 
+  if (!shouldForceAudit()) {
+    return resolve(
+      undefined,
+      'a11yAudit is disabled. To enable, please use the `enableA11yAudit` query param.'
+    );
+  }
```
rwjblue, line 140: I think we should _probably_ remove this change ATM. Explicitly calling `a11yAudit()` should still run the audit "always".

README.md
```diff
@@ -142,17 +142,28 @@ only real difference is Acceptance tests get automatic async handling.
 
 #### Optionally Running a11yAudit
 
-Ember A11y Testing also provides an `a11yAuditIf` helper which will only run
-audits if `enableA11yAudit=true` is set as a query param on the test page. This
-is useful if you want to conditionally run accessibility audits, such as during
-nightly build jobs.
+Ember A11y Testing also allows you to run audits only if `enableA11yAudit=true`
```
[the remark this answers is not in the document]
scalvert: True. Hehe, see what I did there?

README.md
```diff
@@ -142,17 +142,28 @@ only real difference is Acceptance tests get automatic async handling.
...
 test('Some test case', await function(assert) {
   await visit('/');
-  await a11yAuditIf(); // Only runs when enableA11yAudit=true is in the URL
+  await a11yAudit();
+  assert.ok(true, 'no a11y errors found!');
+});
+```
+
+```javascript
+// &enableA11yAudit=false set in the URL
```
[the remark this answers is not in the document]
scalvert: Correct - will remove.

addon-test-support/audit-if.ts
scalvert: That doesn't sound like a user friendly API. I'd prefer to not do this.

addon-test-support/audit.ts
scalvert: You mean until we deprecate? We can, but there isn't really much harm in doing both now, is there?

rwjblue: p5/5 

This is still an issue from my perspective. A few main points:

1. If we only audit when `shouldForceAudit` then, it is **impossible** to implement a `shouldAudit` strategy to "randomly" audit tests. In general, we expect folks that use `shouldAudit` to audit every N test to use the existing `a11yAudit` function as the audit function. In the current implementation of this PR, that **will never** work unless you _also_ have `&enableA11yAudit` query param (which we don't actually intend to be used for anything other than focusing on a test and forcing an audit run). The fact that CI is passing with this change, means we are missing some fairly fundamental coverage in the `shouldAudit` space: we must never actually invoke tests that use `setupGlobalA11yAuditHooks` along with the default `audit` function (if we did, those tests would fail with this change). 

1. We are indeed actively working on making it so that developers do not have to manually call `a11yAudit()` in their own tests (which is I'm sure what is motivating the change you made to this file). As we move forward, users that currently have manual `a11yAudit()` calls will _**loose**_ A11Y coverage if they choose any `shouldAudit` strategy other than `return true`, this is because even the manual testing they had previously done will no longer be ran.

1. As we migrate forward we still need a fundamental internal API that will reliably return the audit results. That API is what the rest of the system should be built on top of, and should be allowed to be used by consumers to compose their own A11y audit solutions. If the fundamental "do an audit and give me results back" API can be completely nooped, we have broken the ability for consumers to "build higher" on top of our building blocks...

addon-test-support/setup-global-a11y-hooks.ts
```diff
@@ -18,7 +17,7 @@ export function setupGlobalA11yHooks(
 ) {
   ['visit', 'click', 'doubleClick', 'tap'].forEach((helperName) => {
     let hook = _registerHook(helperName, 'end', async () => {
-      if (shouldForceAudit() && shouldAudit(helperName, 'end')) {
```
rwjblue, line 21: IMHO, this should be reverted. See https://github.com/ember-a11y/ember-a11y-testing/pull/203/files#r480213736 for details.

addon-test-support/audit.ts
scalvert: Good points. I wonder if we should look to instead provide the `shouldForceAudit` as a compositional primitive, and leave `a11yAudit` as is. We would then simply deprecate the existing API. This would allow people to build in the audit functionality with their invocation strategy to the `setupGlobalA11yAuditHooks`, but would preserve the existing functionality of `a11yAudit`. The only breaking change would then be the ultimate removal of the `a11yAuditIf`. WDYT?

rwjblue: Sounds good to me, that was what I was trying to say over in https://github.com/ember-a11y/ember-a11y-testing/pull/203#discussion_r479532099 but I said the method wrong (and didn't explain what I meant very well).

scalvert: Got it. And we're on the same page. I'll update accordingly.

addon-test-support/audit-if.ts
scalvert: We spoke offline and have agreed that this addon will provide the primitives for users to build and invoke their own `a11yAuditIf`-like functionality.

addon-test-support/setup-global-a11y-hooks.ts
scalvert: Done, as per our discussion.

PR #204, opened 2020-08-28
Converting utils to TS to fix compile time errors

scalvert:
Converts

- addon/utils/format-violation.js
- addon/utils/violations-helper.js

to TS to avoid issue with compile-time errors due to JS imported into TS.

### DIFF
--- a/addon/utils/format-violation.ts
+++ b/addon/utils/format-violation.ts
@@ -1,4 +1,4 @@
-import { Result } from 'axe-core';
+import type { Result } from 'axe-core';
 
 /**
  * Formats the axe violation for human consumption
--- a/addon/utils/format-violation.ts
+++ b/addon/utils/format-violation.ts
@@ -1,10 +1,15 @@
+import { Result } from 'axe-core';
+
 /**
  * Formats the axe violation for human consumption
  *
  * @param {AxeViolation} violation
  * @param {string | string[]} markup (optional) string of HTML relevant to the violation
  */
-export default function formatViolation(violation, markup) {
+export default function formatViolation(
+  violation: Result,
+  markup: string | string[]
+) {
   if (!violation) {
     throw new Error(
       'formatViolation called without required parameter: violation'
--- a/addon/utils/violations-helper.ts
+++ b/addon/utils/violations-helper.ts
@@ -1,3 +1,5 @@
+import { Result } from 'axe-core';
+
 /**
  * @module ember-a11y-testing
  *
@@ -18,6 +20,8 @@
  * @class ViolationsHelper
  */
 export class ViolationsHelper {
+  violations: Result[];
+  hasLoggedTip: boolean;
   /**
    * Instantiate by calling either:
    *
@@ -66,7 +70,7 @@ export class ViolationsHelper {
    * @public
    * @return {Void}
    */
-  push(violation) {
+  push(violation: Result) {
     this.violations.push(violation);
   }
 
@@ -81,13 +85,13 @@ export class ViolationsHelper {
    * @public
    * @return {Array}
    */
-  filterBy(key, value) {
-    if (key === 'rule') {
+  filterBy(key: keyof Result, value: any) {
+    if (<any>key === 'rule') {
       key = 'id';
     }
 
     return this.violations.filter((violation) => {
-      if (key === 'node') {
+      if (<any>key === 'node') {
         return violation.nodes[0].target[0] === value;
       }
       return violation[key] === value;
--- a/addon/utils/violations-helper.ts
+++ b/addon/utils/violations-helper.ts
@@ -1,4 +1,4 @@
-import { Result } from 'axe-core';
+import type { Result } from 'axe-core';
 
 /**
  * @module ember-a11y-testing


addon/utils/format-violation.ts
```diff
@@ -1,10 +1,15 @@
+import { Result } from 'axe-core';
```
rwjblue, line 1:
```suggestion
import type { Result } from 'axe-core';
```

addon/utils/violations-helper.ts
```diff
@@ -1,3 +1,5 @@
+import { Result } from 'axe-core';
```
rwjblue, line 1:
```suggestion
import type { Result } from 'axe-core';
```

addon/utils/violations-helper.ts
```diff
@@ -81,13 +85,13 @@ export class ViolationsHelper {
    * @public
    * @return {Array}
    */
-  filterBy(key, value) {
-    if (key === 'rule') {
+  filterBy(key: keyof Result, value: any) {
+    if (<any>key === 'rule') {
```
scalvert, line 89: This (existing code) seems wrong. There is no `rule` key on an `axe.Result`, so I'm not really sure what this code was trying to do.

addon/utils/violations-helper.ts
```diff
@@ -81,13 +85,13 @@ export class ViolationsHelper {
...
       key = 'id';
     }
 
     return this.violations.filter((violation) => {
-      if (key === 'node') {
+      if (<any>key === 'node') {
```
scalvert, line 94: This (existing code) seems wrong. There is no `node` key on an `axe.Result`, so I'm not really sure what this code was trying to do.

addon/utils/violations-helper.ts
scalvert: For the curious:

```
This condition will always return 'false' since the types '"description" | "help" | "helpUrl" | "id" | "impact" | "tags" | "nodes"' and '"rule"' have no overlap.ts(2367)
```

drewlee: The `rule` property was replaced by `id` in more recent releases of axe.

scalvert: OK, so if we're on a more recent version, can we just remove this? Why would it be kept if we control the version of `axe-core`?

drewlee: No, this looks like an internal construct, not a direct reference to axe.Result, as this is being referred to in tests.

https://github.com/ember-a11y/ember-a11y-testing/blob/e3ceeb27783b12275b26fa743ac08b685d52f9d6/tests/unit/utils/violations-helper-test.js#L84

https://github.com/ember-a11y/ember-a11y-testing/blob/e3ceeb27783b12275b26fa743ac08b685d52f9d6/tests/unit/utils/violations-helper-test.js#L58

drewlee: Gets aggregated started here: https://github.com/ember-a11y/ember-a11y-testing/blob/e3ceeb27783b12275b26fa743ac08b685d52f9d6/addon-test-support/audit.ts#L42

scalvert: So this is where TS is showing there being rough edges in the API.

https://github.com/ember-a11y/ember-a11y-testing/blob/master/addon-test-support/audit.ts#L50

Shows that a `Result` object is being pushed into this helper. 

Similarly:

https://github.com/ember-a11y/ember-a11y-testing/blob/e254ebad3e23d81384a18bd9b91868770d424dc8/addon/instance-initializers/axe-component.js#L180

Is a of type `axe.Result[]`. I'm not sure why it's being used in a different way elsewhere.

What do you think?

scalvert: OK it looks like this is trying to do two different things in two different places. 

https://github.com/ember-a11y/ember-a11y-testing/blob/e254ebad3e23d81384a18bd9b91868770d424dc8/addon/instance-initializers/axe-component.js#L183

Is pushing `NodeResult` into this helper instead. From what I can tell, either type has a `node` or `rule` property.

scalvert: The answer here may be just to extract these other pieces to a separate addon.

drewlee: I think TS may be a little off. As far as I know, the axe results object has the following basic shape:
```
{
  inapplicable: [],
  passes: [],
  violations: [
    {
       id,
       impact,
       tags: [],
       description,
       help,
       nodes: [
          { }
       ]
    }
  ]
}
```

So what gets pushed into violationsHelper is `results.violations[i]`.

drewlee: And it looks like `node` isn't a member of the violations object, it's just used as a means of looking up the first HTML tag in the nodes array.

https://github.com/ember-a11y/ember-a11y-testing/blob/e3ceeb27783b12275b26fa743ac08b685d52f9d6/addon-test-support/audit.ts#L42

drewlee: Yeah, the way this is done isn't very TypeScript friendly, as `key` can either be a member of the violations object, or it can be `rule`, which is normalized to `id` within this function, and `node`, which gets `violation.nodes[0].target[0]`.

rwjblue:
@drewlee / @scalvert - Where are we at with this one?

addon/utils/violations-helper.ts
scalvert: Ya this seems like this data structure is being overloaded in a very odd way. Again, I think the way forward is to extract the component-based pieces into a separate addon, and clean up the existing APIs.

scalvert:
I think this is good to merge. I just confirmed that the changes haven't caused a backwards incompatible change.

PR #205, opened 2020-08-31
Remove audit components in favor of industry standard tools

scalvert:
As a further evolution of this addon, this PR removes the axe specific components that enable development integration in favor of other tools.

As per the README:

While this addon previously included a number of components that would aid in identifying axe violations during development, those have been deprecated in favor of other, industry standard tools such as:

- [**Lighthouse**](https://developers.google.com/web/tools/lighthouse) - an open-source, automated tool for improving the quality of web pages. You can run it against any web page, public or requiring authentication. It has audits for performance, accessibility, progressive web apps, SEO and more.
  - Homepage: https://developers.google.com/web/tools/lighthouse
- [**Sa11y**](https://ryersondmp.github.io/sa11y/) - an accessibility quality assurance tool that visually highlights common accessibility and usability issues. Geared towards content authors, Sa11y indicates errors or warnings at the source with a simple tooltip on how to fix.
- [axe Chrome extension](https://www.deque.com/axe/browser-extensions/) - a free axe browser extension ideal for development teams to test web applications to help identify and resolve common accessibility issues.


MelSumner:
Lighthouse uses axe-core. Can we talk more about this please? I feel like this PR is changing how this addon works.

scalvert:
@MelSumner sure. What specifically would you like to discuss? What are your concerns?

README.md
```diff
@@ -196,171 +196,12 @@ export function a11yAuditIf(contextSelector, axeOptions) {
...
-ENV['ember-a11y-testing']: {
-  componentOptions: {
-    axeOptions: {
-      rules: {
-        'region': { enabled: true } // enable region rule
-      },
-      checks: {
-        'color-contrast': { options: { noScroll: true } } // disable scrolling of color-contrast check
-      }
-    }
```
drewlee, line 353: We should provide instructions on how to do this sort of thing in `test-helper.js` (assuming that's where it would go).

README.md
```diff
@@ -196,171 +196,12 @@ export function a11yAuditIf(contextSelector, axeOptions) {
...
-#### Using Native aXe Global
-
-Ember A11y Testing injects the `axe` global during development and tests so that
-you can run accessibility audits while developing your application. It also
-provides some Ember-specific tools so that you can integrate accessibility
-checks into your workflow easily.
+- [**Lighthouse**](https://developers.google.com/web/tools/lighthouse) - an open-source, automated tool for improving the quality of web pages. You can run it against any web page, public or requiring authentication. It has audits for performance, accessibility, progressive web apps, SEO and more.
+  - Homepage: https://developers.google.com/web/tools/lighthouse
+- [**Sa11y**](https://ryersondmp.github.io/sa11y/) - an accessibility quality assurance tool that visually highlights common accessibility and usability issues. Geared towards content authors, Sa11y indicates errors or warnings at the source with a simple tooltip on how to fix.
+- [axe Chrome extension](https://www.deque.com/axe/browser-extensions/) - a free axe browser extension ideal for development teams to test web applications to help identify and resolve common accessibility issues.
```
drewlee, line 204:
```suggestion
- [**axe Chrome extension**](https://www.deque.com/axe/browser-extensions/) - a free axe browser extension ideal for development teams to test web applications to help identify and resolve common accessibility issues.
```

README.md
```diff
@@ -196,171 +196,12 @@ export function a11yAuditIf(contextSelector, axeOptions) {
...
```
drewlee, line 205: We should also reference Microsoft's Accessibility Insights extension, which is the one we primarily use. I'd probably list it first.
```suggestion
[**Accessibility Insights for Web**](https://accessibilityinsights.io/docs/en/web/overview) - Accessibility Insights for Web helps developers find and fix accessibility issues in web apps and sites. This browser extension for Chrome and the new Microsoft Edge runs on Windows, MacOS, and Linux computers.

```

rwjblue:
> Lighthouse uses axe-core.

Yes, indeed it does, that's why this is a good suggested replacement 😸.

> I feel like this PR is changing how this addon works.

This addon will become focused on testing only (as its name implies). It has been primarily focused on the testing aspects since its inception, and the introduction of the non-testing APIs adds non-trivial complexity that would be better served by a separate dedicated tool (either a separate addon, or usage of a tool that is not an Ember specific one).

README.md
rwjblue: Ya, that's great to call out.

README.md
rwjblue: Ya, it would basically be:

```js
// tests/test-helper.js

// ...snip...
import { setRunOptions } from 'ember-a11y-testing';

setRunOptions({
  rules: {
    region: { enabled: true },
  },
  checks: {
    'color-contrast': {
      options: {
        noScroll: true,
      },
    },
  },
});
```

README.md
```diff
@@ -196,171 +196,12 @@ export function a11yAuditIf(contextSelector, axeOptions) {
 
 ### Development Usage
 
-Usage in development is restricted to applications using Ember 1.13 and up, as it
-relies on the `didRender` hook of a component's life-cycle (a feature only
-available in versions of Ember with the Glimmer rendering engine).
+While this addon previously included a number of components that would aid in identifying axe violations during development, those have been deprecated in favor of other, industry standard tools such as:
```
rwjblue, line 199: Let's reword this to be focused on the go forward users, not about migration to 4.0.0. We should author a custom transition guide for migrating from 3.x to 4.x and we can list the concrete things for those users in that guide.

drewlee:
And probably nix the configuration here: https://github.com/ember-a11y/ember-a11y-testing/blob/c375436e166b61f0d2b158b71ec13ea37fab60db/tests/dummy/config/environment.js#L25

drewlee:
We'll need to revamp the index page of the dummy app as it basically showcases the visual audit component feature.

README.md
scalvert: Agree. Thanks for the callout, @drewlee!

README.md
scalvert: Ah yes. Agree. This was weird as there was an odd interleaving of the documentation for the test-based code and the component-based code. I would have chosen to keep them very separate, but oh well.

I can update the README with this example that @rwjblue showed.

README.md
scalvert: Sounds good.

scalvert:
@drewlee I think as it pertains to the dummy apps' index route, we could just leave it as the acceptance tests use those elements. I think it's reasonable to remove any elements that aren't needed. WDYT?

drewlee:
@scalvert Yeah, that's reasonable. I was thinking the revamp as something we can gradually work thru as opposed to something that needs to be done immediately. Thanks!

README.md
```diff
@@ -194,6 +194,90 @@ export function a11yAuditIf(contextSelector, axeOptions) {
...
+```
+
+Test module level:
+
+```javascript
+import { module, test } from 'qunit';
+import { setRunOptions } from 'ember-a11y-testing';
+
+module('some test module', function (hooks) {
+  hooks.before(function () {
```
rwjblue, line 224: Needs to be `beforeEach` (this is because we reset after each test):

```suggestion
  hooks.beforeEach(function () {
```

README.md
```diff
@@ -194,6 +194,90 @@ export function a11yAuditIf(contextSelector, axeOptions) {
...
+    // ...
+    a11yAudit();
+    // ...
+  });
+
+  // ...
+});
+```
+
+Options are automatically torn down when the test context's owner is destroyed.
```
rwjblue, line 274:
```suggestion
When using `setRunOptions` during a test, the options you set are automatically reset when the test completes.
```

addon-test-support/audit.ts
```diff
@@ -111,10 +90,10 @@ export function _normalizeRunParams(
   }
 
   if (typeof options !== 'object') {
-    options = config[_configName]?.componentOptions?.axeOptions || {};
+    options = getRunOptions() || {};
```
rwjblue, line 93: Why do we need to default? I think `axe-core` can handle `undefined` for options...

index.js
```diff
@@ -36,7 +34,7 @@ module.exports = {
     );
 
     if (app.tests) {
-      let type = shouldExcludeAxeCore ? { type: 'test' } : undefined;
+      let type = options.excludeAxeCore ? { type: 'test' } : undefined;
```
rwjblue, line 37: I think this option is probably defunct now and should be removed?

README.md
scalvert: Ya I originally was going to do this, but thought it made more sense to do it once rather than for each test invoked. I supposed it doesn't matter, but it's not clear to me what `getContext().owner` would return in `hooks.before`. What would it be?

addon-test-support/audit.ts
scalvert: This was the old behavior, so this was just preserving it.

index.js
scalvert: Ya, I wasn't sure why it was there. I can remove.

addon-test-support/audit.ts
scalvert: I think I can probably remove it, since that param is optional anyway.

scalvert:
Thanks, @drewlee! Really appreciate your careful review. I missed a bunch of things in the first iteration.

PR #212, opened 2020-09-04
Adding ability to post violations via middleware

scalvert:
Adds support for combining the custom reporter with custom middleware, which allows for serializing the results to JSON.

```js
// tests/test-helper.js
import { setupMiddlewareReporter } from 'ember-a11y-testing';
// ...

setupMiddlewareReporter();

// ...
```

This will setup a custom reporter, which gathers the test module name, test name, and violations for each failed audit. Once the suite is done, these results are posted to the testem middleware, ultimately serializing the results to disc.

TODO:
- [x] Add tests for `setupMiddlewareReporter`
- [x] Update `setupMiddleware` test to use real `AxeResults`
- [x] Determine which specific directory path to pass to `setupMiddlewareReporter` in order to save the results.

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,6 +5,7 @@ const fs = require('fs');
 const Funnel = require('broccoli-funnel');
 const MergeTrees = require('broccoli-merge-trees');
 const VersionChecker = require('ember-cli-version-checker');
+const setupMiddleware = require('./setup-middleware');
 
 // The different types/area for which we have content for.
 const ALLOWED_CONTENT_FOR = ['head-footer', 'test-head-footer'];
@@ -60,5 +61,17 @@ module.exports = {
         path.join(__dirname, 'content-for', type + '.html')
       );
     }
+  },
+
+  serverMiddleware(startOptions) {
+    setupMiddleware(startOptions.app, {
+      root: this.project.root,
+    });
+  },
+
+  testemMiddleware(app) {
+    setupMiddleware(app, {
+      root: this.project.root,
+    });
   },
 };
--- a/node-tests/fixtures/violations.json
+++ b/node-tests/fixtures/violations.json
@@ -1,0 +1,3325 @@
+{
+  "testEngine": {
+    "name": "axe-core",
+    "version": "4.0.1"
+  },
+  "testRunner": {
+    "name": "axe"
+  },
+  "testEnvironment": {
+    "userAgent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/85.0.4183.83 Safari/537.36",
+    "windowWidth": 1390,
+    "windowHeight": 1089,
+    "orientationAngle": 0,
+    "orientationType": "landscape-primary"
+  },
+  "timestamp": "2020-09-04T18:11:29.913Z",
+  "url": "http://localhost:4200/tests/index.html?testId=40d80690&enableA11yAudit=",
+  "toolOptions": {
+    "reporter": "v1"
+  },
+  "violations": [
+    {
+      "id": "button-name",
+      "impact": "critical",
+      "tags": [
+        "cat.name-role-value",
+        "wcag2a",
+        "wcag412",
+        "section508",
+        "section508.22.a"
+      ],
+      "description": "Ensures buttons have discernible text",
+      "help": "Buttons must have discernible text",
+      "helpUrl": "https://dequeuniversity.com/rules/axe/4.0/button-name?application=axeAPI",
+      "nodes": [
+        {
+          "any": [
+            {
+              "id": "button-has-visible-text",
+              "data": null,
+              "relatedNodes": [],
+              "impact": "critical",
+              "message": "Element does not have inner text that is visible to screen readers"
+            },
+            {
+              "id": "aria-label",
+              "data": null,
+              "relatedNodes": [],
+              "impact": "serious",
+              "message": "aria-label attribute does not exist or is empty"
+            },
+            {
+              "id": "aria-labelledby",
+              "data": null,
+              "relatedNodes": [],
+              "impact": "serious",
+              "message": "aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty"
+            },
+            {
+              "id": "role-presentation",
+              "data": null,
+              "relatedNodes": [],
+              "impact": "minor",
+              "message": "Element's default semantics were not overridden with role=\"presentation\""
+            },
+            {
+              "id": "role-none",
+              "data": null,
+              "relatedNodes": [],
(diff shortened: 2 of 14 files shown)


node-tests/fixtures/violations.json
```diff
@@ -0,0 +1,3325 @@
+{
```
rwjblue, line 1: I think we need to tweak the format into something akin to:

```js
[
  {
    testName: 'asdf',
    results: [
      {
        helperName: 'click',
        stack: '',
        label: 'end',
        axeResult: WhateverFromAxeHere
      }    
    ]
  }
]
```

package.json
```diff
@@ -23,19 +23,27 @@
...
     "ember-cli-babel": "^7.21.0",
     "ember-cli-typescript": "^4.0.0",
     "ember-cli-version-checker": "^5.1.1",
-    "ember-destroyable-polyfill": "^2.0.1"
+    "ember-destroyable-polyfill": "^2.0.1",
+    "express": "^4.17.1",
+    "fs-extra": "^9.0.1",
+    "node-fetch": "^2.6.0",
+    "qunit": "^2.11.0",
+    "tmp": "^0.2.1"
```
rwjblue, line 46: I think these might need to be devDependencies?

setup-middleware.js
```diff
@@ -0,0 +1,36 @@
...
+
+const bodyParser = require('body-parser').json({ limit: '50mb' });
+const writeJsonSync = require('fs-extra').writeJSONSync;
+const path = require('path');
+const date = require('date-and-time');
+
+function reportViolations(req, res, options) {
+  const REPORT_TIMESTAMP = date.format(new Date(), 'YYYY-MM-DD-HH_mm_ss');
+  let outputPath = path.resolve(
+    path.join(options.root, `ember-a11y-report-${REPORT_TIMESTAMP}.json`)
```
rwjblue, line 11: Maybe: 

```js
    path.join(options.root, 'ember-a11y-report', `${REPORT_TIMESTAMP}.json`)
```

rwjblue (changes requested):

node-tests/fixtures/violations.json
scalvert: Ya I can explore doing something like what we did for test isolation validation in the capturing of the stack. I think we can use the `testMetaData` feature we added to `@ember/test-helpers` to get the helpers that were invoked, the last being the one that triggered the audit.

package.json
scalvert: Yep, I was avoiding the `no-unpublished-require` lint violation, but as you pointed out just adding that to `.npmignore` is the right path here.

node-tests/fixtures/violations.json
rwjblue: Ya, if we can use the test meta data stuff we might even avoid needing to use QUnit itself to grab the info?

drewlee:
Just as a general question, since I didn't see this directly referred to in the code - how does the host application determine whether they want to run tests w/reporter vs no reporter (i.e., conditionally)? For example, we may not need the reporter for CI/CD process or local execution, while requiring the reporter for a nightly job. Would we just wrap `setupMiddlewareReporter()` in a conditional check in `tests/test-helper.js`?

node-tests/fixtures/violations.json
scalvert: So `results` wouldn't be an array. It'd be an `AxeResults` instance, which is an object that looks like [this](https://raw.githubusercontent.com/scalvert/ember-a11y-testing/reporter-middleware/node-tests/fixtures/violations.json), as you referenced. That's the raw `AxeResults` value we get back from an individual run. This would just be an object.

I'd propose we do the following:

```json
{
  testName: 'asdf',
  result: 
  {
    helperName: 'click',
    stack: '',
    axeResult: WhateverFromAxeHere
  }    
}
```

I'm not sure what value the label would provide, so I've removed it. Also, if we wanted that info, we'd need to thread it through the hooks invocation, but that's not guaranteed to be present, as you can invoke `a11yAudit` on its own with this custom middleware.

scalvert:
@drewlee - yes. I think we can configure that in the test-helper.js file via a custom query param.

scalvert:
@rwjblue WDYT about us moving the parallel execution support (`ember-exam`) work to a subsequent PR? This will allow us to ship this increment, and be able to build out the subsequent supporting parts later.

setup-middleware.js
```diff
@@ -0,0 +1,39 @@
...
+
+function reportViolations(req, res, options) {
+  const REPORT_TIMESTAMP = date.format(new Date(), 'YYYY-MM-DD-HH_mm_ss');
+  let outputDir = path.join(options.root, 'ember-a11y-report');
+  let outputPath = path.resolve(
+    path.join(outputDir, `${REPORT_TIMESTAMP}.json`)
+  );
+
+  ensureDirSync(outputDir);
+  writeJsonSync(outputPath, req.body);
```
rwjblue, line 16: I _think_ these two calls are redundant, right? Doesn't `writeJsonSync` internally `ensureDirSync`?

node-tests/setup-middleware-test.js
```diff
@@ -0,0 +1,59 @@
...
+  let { module, testName } = QUnit.config.current;
+
+  return {
+    moduleName: module.name,
+    testName,
+    helperName: 'visit',
+    stack: 'STACK',
+    axeResults,
+  };
+}
```
rwjblue, line 24: I think this is still not quite right (we don't want the JSON file to be one per a11yAudit invocation). I'm happy to iterate in separate PRs, but lets make an issue to chat through the goal/intent here (and the final format).

setup-middleware.js
scalvert: It does not.

node-tests/setup-middleware-test.js
scalvert: Are you saying the test isn't right? The test doesn't mimic exactly how this will be wired up IRL, but merely tests the emitting and gathering of the data. 

Can you be more specific?

index.js
```diff
@@ -61,4 +62,16 @@ module.exports = {
...
+    setupMiddleware(startOptions.app, {
+      root: this.project.root,
+    });
+  },
+
+  testemMiddleware(app) {
+    setupMiddleware(app, {
+      root: this.project.root,
+    });
+  },
```
rwjblue, line 76: The differences in these methods (from ember-cli's side) is very annoying 😭

setup-middleware.js
rwjblue: Whoops! Sorry, I was thinking of `outputJsonSync` 

https://github.com/jprichardson/node-fs-extra/blob/master/docs/outputJson-sync.md

rwjblue (approved):

PR #214, opened 2020-09-09
Adding enhanced console logging

scalvert:
Recent refactors removed the feature that output failures to the console. This PR restores that, using a slightly nicer format inspired by https://github.com/dequelabs/react-axe. 

[An issue](https://github.com/dequelabs/react-axe/issues/180) has been opened on that repository which proposes extracting the formatting functionality to a separate repo, in order to reduce duplication.

Example:

Screen Shot 2020-09-09 at 9 36 57 AM

### DIFF
--- a/addon-test-support/index.ts
+++ b/addon-test-support/index.ts
@@ -13,5 +13,6 @@ export {
   setupMiddlewareReporter,
 } from './setup-middleware-reporter';
 export { storeResults, printResults } from './logger';
+export { setupConsoleLogger } from './setup-console-logger';
 
 export { InvocationStrategy, A11yAuditReporter } from './types';
--- a/addon-test-support/index.ts
+++ b/addon-test-support/index.ts
@@ -12,5 +12,6 @@ export {
   middlewareReporter as _middlewareReporter,
   setupMiddlewareReporter,
 } from './setup-middleware-reporter';
+export { storeResults, printResults } from './logger';
 
 export { InvocationStrategy, A11yAuditReporter } from './types';
--- a/addon-test-support/logger.ts
+++ b/addon-test-support/logger.ts
@@ -1,7 +1,7 @@
 import axeCore, { AxeResults, NodeResult, RelatedNode, Result } from 'axe-core';
 
 /**
- * This file is heavily borrowed from https://github.com/dequelabs/react-axe
+ * This file is heavily borrowed from https://github.com/dequelabs/react-axe/blob/d3245b32fc5ed19e3c7b2c43d2815fe63f5875cb/index.ts
  *
  * An issue (https://github.com/dequelabs/react-axe/issues/180) has been opened which
  * suggests extracting out the common pieces of axe formatting to a separate repository,
--- a/addon-test-support/logger.ts
+++ b/addon-test-support/logger.ts
@@ -1,0 +1,186 @@
+import axeCore, { AxeResults, NodeResult, RelatedNode, Result } from 'axe-core';
+
+/**
+ * This file is heavily borrowed from https://github.com/dequelabs/react-axe
+ *
+ * An issue (https://github.com/dequelabs/react-axe/issues/180) has been opened which
+ * suggests extracting out the common pieces of axe formatting to a separate repository,
+ * which would allow a lot of these functions to be removed.
+ */
+
+// @reference axe-core https://github.com/dequelabs/axe-core/blob/develop/lib/core/base/audit.js
+type AxeCoreNodeResultKey = 'any' | 'all' | 'none';
+
+interface AxeWithAudit {
+  _audit: {
+    data: {
+      failureSummaries: {
+        any: {
+          failureMessage: (args: string[]) => string;
+        };
+        all: {
+          failureMessage: (args: string[]) => string;
+        };
+        none: {
+          failureMessage: (args: string[]) => string;
+        };
+      };
+    };
+  };
+}
+
+// contrasted against Chrome default color of #ffffff
+const lightTheme = {
+  serious: '#d93251',
+  minor: '#d24700',
+  text: 'black',
+};
+
+// contrasted against Safari dark mode color of #535353
+const darkTheme = {
+  serious: '#ffb3b3',
+  minor: '#ffd500',
+  text: 'white',
+};
+
+const theme =
+  window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches
+    ? darkTheme
+    : lightTheme;
+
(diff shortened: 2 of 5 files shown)


addon-test-support/logger.ts
```diff
@@ -0,0 +1,186 @@
+import axeCore, { AxeResults, NodeResult, RelatedNode, Result } from 'axe-core';
+
+/**
+ * This file is heavily borrowed from https://github.com/dequelabs/react-axe
```
rwjblue, line 4: Can you make this a permalink to the actual source (so we can go to that location to track changes over time)?

addon-test-support/logger.ts
```diff
@@ -0,0 +1,186 @@
...
+  const el = document.querySelector(node.target.toString());
+  if (!el) {
+    logFn('Selector: %c%s', boldCourier, node.target.toString());
+  } else {
+    logFn('Element: %o', el);
+  }
+}
+
+/**
+ * Log the axe result node html tot he console
```
rwjblue, line 91:
```suggestion
 * Log the axe result node html to the console
```

rwjblue (commented):
Remind me, how does `console.group` and friends work in CI? Does `testem` ignore them?

tests/test-helper.js
```diff
@@ -1,8 +1,14 @@
...
 import config from 'dummy/config/environment';
 import { setApplication } from '@ember/test-helpers';
 import { start } from 'ember-qunit';
+import { printResults } from 'ember-a11y-testing/test-support';
 
 setApplication(Application.create(config.APP));
 
+QUnit.done(function () {
+  printResults();
+});
```
drewlee, line 12: I'm assuming that the host application would have to do a similar setup to get console output? Is there any benefit to doing it this way as opposed to just getting it for free when using the default reporter? This way we preserve original functionality without introducing extra setup work.

rwjblue: @drewlee - Ya, the integration within the `DEFAULT_REPORTER` just above, should ensure that all applications gather the output for the final summary but you are right they won't actually log out (with the current set of changes).

drewlee: Oh, I see what's up. We're now outputting all of the results to the console at the end of the test suite, as opposed to each individual test. I think we'll need to get this documented somewhere as there's likely going to be a lot of people wondering WTF results aren't being output in the console anymore by default.

addon-test-support/logger.ts
scalvert: Ha. I guess they need to fix this in their repo too :)

addon-test-support/logger.ts
scalvert: Sure.

tests/test-helper.js
scalvert: So we removed the old `console.error` calls. They were hard to read, and were interspersed throughout the console output. If you had deprecations, other errors, info, warnings, etc, you'd have a really hard time finding the complete list of violations. This way we avoid that by outputting them all in one `console.group`.

As for not wiring it up automatically, I wanted to avoid tightly coupling our code to `QUnit`, but I can make it work for both `QUnit` and `mocha`, the two currently supported test frameworks. I'd still like to provide a setup function that you explicitly call to get the console output, rather than trying to bind it automatically, since this allows those who don't want the console noise to just not setup the console output. WDYT?

drewlee: Understood from a technical perspective. But from a developer experience perspective, it's a really useful feature that seems strange not to have as a default, especially when it makes debugging the source of failures so much easier. I'm not sure if there would be a valid reason for a consumer not to want it. Personally, I'd prefer to avoid "configuration overhead" for core features (meaning that previously, they just worked out of the box), but ultimately I'll leave that decision up to you. 🙂

scalvert: OK, I added a `setupConsoleLogger` function that can be imported and run in the test-helper.js file, which will bind the output to render on `QUnit.done`. I'm going to omit adding an implementation for mocha, as mocha's root hooks are configured in a different way, and I'd rather leave the functional primitives for app authors to implement on their own.

We'll want to add this direction to the documentation, which I plan on adding at the end of this range of refactorings before we publish.

addon-test-support/index.ts
```diff
@@ -12,5 +12,7 @@ export {
   middlewareReporter as _middlewareReporter,
   setupMiddlewareReporter,
 } from './setup-middleware-reporter';
+export { storeResults, printResults } from './logger';
+export { setupConsoleLogger } from './setup-console-logger';
```
rwjblue, line 16: Just a note: we should queue up a subsequent conversation RE: providing a generic `import { setup } from 'ember-a11y-testing'` type API that internally calls all of the methods we are exporting. My idea here is to make it roughly similar to the things we are doing in ember-qunit's `start` (where each thing it does can be opted out of, and controlled manually by dropping down a layer as needed).

https://github.com/emberjs/ember-qunit/blob/1d3decdd60b78bc07a4bc4c20f53075de6467600/addon-test-support/index.js#L175-L230

AFAICT, we have all the required pieces, and it would be easy to add in a follow up.

addon-test-support/setup-console-logger.ts
```diff
@@ -0,0 +1,8 @@
+import QUnit from 'qunit';
+import { printResults } from './logger';
+
+export function setupConsoleLogger() {
```
rwjblue, line 4: I think we'll need to add some docs to the README for this (and possibly some API docs).

addon-test-support/index.ts
scalvert: Yep, I like this idea, and agree.

addon-test-support/setup-console-logger.ts
scalvert: I mentioned this in [my comment above](https://github.com/ember-a11y/ember-a11y-testing/pull/214#discussion_r485843742). I'm going to do a full docs pass on the entire repo once all these changes are merged in.

rwjblue: Sounds good, mind double checking the meta issue to make sure we have this logged?

scalvert: For sure, and it is!

scalvert: I updated the line item to be a bit more precise to what we're needing, namely a README update for the API and functional changes.

scalvert:
> Remind me, how does `console.group` and friends work in CI? Does `testem` ignore them?

I believe so, yes. Good luck finding where the console hijacking is in testem though!

drewlee (approved):

PR #219, opened 2020-09-16
Update ember-qunit to latest beta.

rwjblue:
Fixes some issues with apps/addons that do not have ember-auto-import themselves.


package.json
```diff
@@ -90,7 +90,6 @@
     "npm-run-all": "^4.1.5",
     "prettier": "^2.0.5",
     "qunit": "^2.11.0",
-    "qunit-dom": "^1.4.0",
```
rwjblue, line 93: qunit-dom is not compatible (yet, working on a PR) with the latest ember-qunit beta, and since we don't directly use it I figured I'd just remove it...

scalvert: Makes sense.

scalvert (approved):

PR #233, opened 2021-02-04
Fix cannot find module `qunit` under embroider

thoov:
When parent applications are using ember-qunit > 5 this causes an invalid lookup for qunit in requirejs because ember-qunit no longer provides the AMD shim for qunit and instead uses ember-auto-import. This change conditionally uses ember-auto-import if that shim is present or not along with adding test cases for general embroider compatibility.

Related issue: https://github.com/ember-cli/ember-exam/issues/634

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -16,17 +16,11 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/setup-node@v1
-        with:
-          node-version: 10.x
-      - name: install dependencies
-        run: yarn install --frozen-lockfile --non-interactive
-      - name: lint
-        run: yarn lint
-      - name: test
-        run: yarn test:ember
-      - name: test node
-        run: yarn test:node
+      - uses: volta-cli/action@v1
+      - run: yarn install --frozen-lockfile
+      - run: yarn lint
+      - run: yarn test:ember
+      - run: yarn test:node
 
   floating-dependencies:
     name: Floating dependencies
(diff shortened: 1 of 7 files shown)


thoov:
cc: @rwjblue

ember-cli-build.js
```diff
@@ -3,8 +3,11 @@
 const EmberAddon = require('ember-cli/lib/broccoli/ember-addon');
 
 module.exports = function (defaults) {
+  const self = defaults.project.findAddonByName('ember-a11y-testing');
+  const autoImport = self.options.autoImport;
   let app = new EmberAddon(defaults, {
     // Add options here
```
rwjblue, line 9: Mostly just doing this to trigger a CI build 😝 
```suggestion
```

rwjblue (approved):

rwjblue:
Rebased on top of #234

.github/workflows/ci-build.yml
```diff
@@ -16,37 +16,26 @@ jobs:
...
-      - uses: actions/setup-node@v1
-        with:
-          node-version: 10.x
-      - name: install dependencies
-        run: yarn install --no-lockfile --non-interactive
-      - name: test
-        run: yarn test:ember-compatibility
+      - uses: volta-cli/action@v1
+      - run: yarn install --no-lockfile
+      - run: yarn test:ember-compatibility
```
rwjblue, line 33:
```suggestion
      - run: yarn test:ember
      - run: yarn test:node
```

We shouldn't run all Ember version compat checks (those are already ran with `--no-lockfile` in their individual jobs)

PR #239, opened 2021-02-25
Correct import path of setRunOptions in README

danwenzel:
setRunOptions is imported from /test-support, not the root


rwjblue:
Thank you!

PR #243, opened 2021-03-04
Correct link to axe-core API docs

danwenzel:
Updates the URL to the axe-core API docs.  The previous link, while it properly redirected the path, brought you to the top of the page (the element id linked by the hash must have changed at some point)


rwjblue (approved):
Thanks!

PR #247, opened 2021-04-05
Fixes CI build failures for missing module 'levenary'

drewlee:
Builds in CI are failing due to "Cannot find module 'levenary'". Looks like the new releases of `ember-cli-babel` and `ember-auto-import` no longer use this transient dependency, so the fix upgrades these. This makes the dependency resolutions in sync with using `yarn install --frozen-lockfile`.


drewlee:
Still seeing build failures with ember-beta and ember-canary, but these look completely unrelated.

Issue #248, opened 2021-04-06
Axe-core double inclusion via Webpack


drewlee:
Host apps are including `axe-core` duplicate times, once via the standard `app.import` method set up by the addon, and the second time via Webpack, which is presumably from `ember-auto-import`. Both appear in the test-support.js bundle. There is also associated log output when building the host app:

```
WARNING in ./node_modules/axe-core/axe.js 11:149698-149705
Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
 @ /private/var/folders/c5/w1hrxnv55sx_303npp2r5q_h000510/T/broccoli-29237IvTtzxKWMRSR/cache-2222-bundler/staging/tests.js
 @ multi /private/var/folders/c5/w1hrxnv55sx_303npp2r5q_h000510/T/broccoli-29237IvTtzxKWMRSR/cache-2222-bundler/staging/l.js /private/var/folders/c5/w1hrxnv55sx_303npp2r5q_h000510/T/broccoli-29237IvTtzxKWMRSR/cache-2222-bundler/staging/tests.js

WARNING in ./node_modules/axe-core/axe.js
Module not found: Error: Can't resolve 'crypto' in '/Users/andlee/workspace/lighthouse-web_trunk/node_modules/axe-core'
 @ ./node_modules/axe-core/axe.js
```

Standard Inclusion
<img width="1862" alt="Screen Shot 2021-04-06 at 12 51 15 PM" src="https://user-images.githubusercontent.com/945868/113771091-d0ca8380-96d7-11eb-814f-50095a8e2c34.png">

Webpack Injection
<img width="1862" alt="Screen Shot 2021-04-06 at 12 51 30 PM" src="https://user-images.githubusercontent.com/945868/113771137-de800900-96d7-11eb-9b5f-6b03b9b670d1.png">

Not sure if it's unique to a certain set of dependencies, but the host app includes:
- ember-cli 3.16.2
- ember-source 3.16.9
- ember-auto-import 1.11.2

PR #249, opened 2021-04-06
Fixes duplicate axe-core inclusions in host apps

drewlee:
Fixes https://github.com/ember-a11y/ember-a11y-testing/issues/248. Verified working with a couple of host apps, but I'm not sure if there are any other subtleties or nuances that need to be covered.

### DIFF
--- a/index.js
+++ b/index.js
@@ -27,6 +27,9 @@ module.exports = {
       .for('ember-qunit')
       .lt('5.0.0-beta.1');
 
+    // Prevent duplicate axe-core inclusions in host application
+    const exclude = ['axe-core'];
+
     // Ember-qunit < 5 provides an AMD shim for qunit but newer version now use
     // ember-auto-import to include qunit. This means that qunit is no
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -27,6 +27,9 @@ module.exports = {
       .for('ember-qunit')
       .lt('5.0.0-beta.1');
```
scalvert, line 28: Should we also update this to point to the full major?

scalvert (changes requested):

index.js
rwjblue: The main reason we used the beta was because this was the first version that used the new setup, so I think `5.0.0-beta.1` is still the correct value.

rwjblue (approved):

index.js
scalvert: Sure, it just seems odd to have beta be the value we test for, though technically it's correct.

scalvert (approved):

drewlee:
Dropping this in favor of PR #251.

PR #251, opened 2021-04-13
Removes unnecessary, legacy build code for inclusion of axe dependencies.

### LINKED ISSUE #248: Axe-core double inclusion via Webpack

MelSumner:
should fix #248

- checked with @ef4 & the lines removed are per his recommendation (as the author of `ember-auto-import`) - seems we were doing some redundant things that `ember-auto-import` was already providing. 
- updates the version of `ember-qunit`

I did notice that some of the dummy app tests are failing. Perhaps we should plan to revisit the app and update it. I'll make a new issue for that.

### DIFF
--- a/index.js
+++ b/index.js
@@ -23,13 +23,13 @@ module.exports = {
 
     const hasMagicallyProvidedQUnit = versionChecker
       .for('ember-qunit')
-      .lt('5.1.4');
+      .lt('5.0.0-beta.1');
 
     // Ember-qunit < 5 provides an AMD shim for qunit but newer version now use
     // ember-auto-import to include qunit. This means that qunit is no
     // longer available for addons (if the parent app is using ember-qunit > 5) to
     // directly import under embroider unless they are using ember-auto-import
-    // themselves. This condidionally falls back to not using ember-auto-import
+    // themselves. This conditionally falls back to not using ember-auto-import
     // when the parent app is providing qunit because without this we would double
     // include qunit resulting in a runtime error (qunit detects if it as
     // already be added to the window object and errors if so).
--- a/index.js
+++ b/index.js
@@ -25,6 +25,15 @@ module.exports = {
       .for('ember-qunit')
       .lt('5.0.0-beta.1');
 
+    this.options = this.options || {};
+    this.options.autoImport = {
+      webpack: {
+        module: {
+          noParse: /\baxe\.js$/,
+        },
+      },
+    };
+
     // Ember-qunit < 5 provides an AMD shim for qunit but newer version now use
     // ember-auto-import to include qunit. This means that qunit is no
     // longer available for addons (if the parent app is using ember-qunit > 5) to
@@ -34,10 +43,7 @@ module.exports = {
     // include qunit resulting in a runtime error (qunit detects if it as
     // already be added to the window object and errors if so).
     if (hasMagicallyProvidedQUnit) {
-      this.options = this.options || {};
-      this.options.autoImport = {
-        exclude: ['qunit'],
-      };
+      this.options.autoImport.exclude = ['qunit'];
     }
   },
 
--- a/index.js
+++ b/index.js
@@ -2,8 +2,6 @@
 
 const path = require('path');
 const fs = require('fs');
-const Funnel = require('broccoli-funnel');
-const MergeTrees = require('broccoli-merge-trees');
 const VersionChecker = require('ember-cli-version-checker');
 const validatePeerDependencies = require('validate-peer-dependencies');
 const setupMiddleware = require('./setup-middleware');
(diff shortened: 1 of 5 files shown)


index.js
```diff
@@ -25,7 +23,7 @@ module.exports = {
 
     const hasMagicallyProvidedQUnit = versionChecker
       .for('ember-qunit')
-      .lt('5.0.0-beta.1');
+      .lt('5.1.4');
```
rwjblue, line 26: I think this part of the change should be reverted, see my comment on #249 for details

MelSumner: fixed! 👍

scalvert (approved):
I'm unsure myself as to whether this is 100% required or not, but I'll defer to @ef4 to determine the correct path here.

ef4 (approved):
This looks good to me. But there are some things left behind that are no longer used after this PR:

 - https://github.com/ember-a11y/ember-a11y-testing/blob/master/vendor/shims/axe-core.js 
 - broccoli-funnel in package.json
 - broccoli-merge-trees in package.json

MelSumner:
> This looks good to me. But there are some things left behind that are no longer used after this PR:
> 
> * https://github.com/ember-a11y/ember-a11y-testing/blob/master/vendor/shims/axe-core.js
> * broccoli-funnel in package.json
> * broccoli-merge-trees in package.json

ahh good catch, I'll fix that. Thank you, @ef4 !

scalvert:
The beta and canary CI tests are failing in this PR. It looks like they're also failing in master, so we'll need to figure it out.

drewlee:
Have we ran these with a bunch of host apps just to be sure this works as expected? My main concerns are
  1. Axe-core isn't inadvertently being included in non-test builds,
  2. The following build errors need to be removed which seems to come from Webpack or some other entity in ember-auto-import: ```
 WARNING in ./node_modules/axe-core/axe.js 11:149698-149705
Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
 @ /private/var/folders/c5/w1hrxnv55sx_303npp2r5q_h000510/T/broccoli-29237IvTtzxKWMRSR/cache-2222-bundler/staging/tests.js
 @ multi /private/var/folders/c5/w1hrxnv55sx_303npp2r5q_h000510/T/broccoli-29237IvTtzxKWMRSR/cache-2222-bundler/staging/l.js /private/var/folders/c5/w1hrxnv55sx_303npp2r5q_h000510/T/broccoli-29237IvTtzxKWMRSR/cache-2222-bundler/staging/tests.js

WARNING in ./node_modules/axe-core/axe.js
Module not found: Error: Can't resolve 'crypto' in '/Users/andlee/workspace/lighthouse-web_trunk/node_modules/axe-core'
 @ ./node_modules/axe-core/axe.js
```

**Comment by MelSumner** [comment]
@drewlee 

1. I haven't tried this with any host apps yet, I expected the tests in the addons to provide appropriate coverage
2. I'll look into these build errors to see if I can track it down, but if @ef4 has more context in the immediate, please weigh in. 

Thanks for the feedback!

**Comment by ef4** [comment]
It looks like axe-core is trying to fix that warning upstream. It's harmless but annoying. You can avoid it like this:

```js
autoImport: {
  webpack: {
    module: {
      noParse: /\baxe\.js$/
    }
  }
}
```

drewlee:
@MelSumner I ran this through a few apps and everything looks good, outside of some messaging related to axe-core being deoptimized due to being over 500k in size. I think the "crypto" issue got fixed in axe-core itself, so we may need to bump it, but I need to verify. Afterwards, I can merge this and publish a new release later today.

drewlee:
Deque hasn't published a release with the fix yet, so we'll need to add Edward's suggestion above.

drewlee:
The "crypto" issue has been documented under https://github.com/dequelabs/axe-core/issues/2873, which I believe should be appropriately handled in axe-core itself.

drewlee (approved):

PR #255, opened 2021-04-16
Removes deprecated visual audit related components

drewlee:
As the visual audit feature has been deprecated, this change removes unnecessary components from the dummy app and updates related files to be in line with Octane conventions. This allows us to upgrade ESLint (PR #230) and Ember Template Lint (PR #241).

Note that I've had to use a couple of empty glimmer component classes to get tests to pass in classic mode and Ember 3.8 & 3.12, as `templateOnly` isn't available in those two versions.


scalvert:
We should consider dropping 3.8 and 3.12, since I believe the latest LTS is 3.16.

scalvert (approved):

Issue #260, opened 2021-05-10
Unnecessary styles injected into host app's index.html from `ember-a11y-testing`


drewlee:
Unnecessary CSS styles are being injected into header content by the `ember-a11y-testing` addon. The styles are an artifact of the addon's deprecated visual component audit feature and should be removed.

```css
<style>
  .axe-violation--level-1:not(:focus):not(:hover) {
    background-image: repeating-linear-gradient(
      135deg,
      transparent,
      transparent 10%,
      hsla(0, 0%, 88%, 1) 10%,
      hsla(0, 0%, 62%, 1) 15%,
      hsla(0, 0%, 38%, 1) 20%,
      transparent 20%,
      transparent 35%
    ) !important;
  }  .axe-violation--level-2:not(:focus):not(:hover) {
    background-image: repeating-linear-gradient(
      135deg,
      transparent,
      transparent 25%,
      /* make transparent portion somewhat wider than total stripe width */
        hsla(60, 79%, 59%, 1) 25%,
      hsla(60, 79%, 59%, 1) 30%,
      hsla(166, 44%, 65%, 1) 30%,
      hsla(166, 44%, 65%, 1) 35%,
      hsla(240, 26%, 50%, 1) 35%,
      hsla(240, 26%, 50%, 1) 40%
    ) !important;
  }  .axe-violation--level-3:not(:focus):not(:hover) {
    background-image: repeating-linear-gradient(
      135deg,
      transparent,
      transparent 0.5em,
      hsla(60, 79%, 59%, 1) 0.5em,
      hsla(60, 79%, 59%, 1) 1em,
      hsla(240, 26%, 50%, 1) 1em,
      hsla(240, 26%, 50%, 1) 1.5em
    ) !important;
  }  .axe-violation--replaced-element:not(:focus):not(:hover) {
    box-shadow: 0 0 0em 0.35em hsla(240, 26%, 50%, 1) inset,
      0 0 0em 0.35em hsla(60, 79%, 59%, 1), 0 0 0em 0.7em hsla(240, 26%, 50%, 1) !important;    /* Scale down slightly so box shadow is visible */
    transition-duration: 0.3s !important;
    transition-timing-function: ease-in-out !important;
    transform: scale(0.9) !important;
  }
</style>
```

drewlee:
Fixed via PR #261

PR #261, opened 2021-05-10
Removes deprecated contet-for styles

drewlee:
Fixes issue #260 by removing the injection for deprecated visual component audit CSS styles.


drewlee:
The Embroider-related checks started failing a couple of weeks ago. Not super obvious to the cause, but since they're unrelated to this change, I think it would better to pursue those via a separate loop.

MelSumner (approved):
Makes sense. Thank you for doing this work! 👍

PR #268, opened 2021-06-08
Adjusting time of data gathering in middleware reporter

scalvert:
There's a subtle timing issue that occurs in the middleware reporter, whereby we can't guarantee that all test metadata is setup by the time it's accessed. The only way to ensure all metadata is setup for the test is by accessing it lazily, via `testDone`.

This will become particularly important when trying to gather additional data destined for the middleware report, such as the filepath work that's [currently being implemented](https://github.com/glnster/ember-babel-add-test-metadata). We'd want to ensure we access this metadata inside `pushTestResult`.


addon-test-support/setup-middleware-reporter.ts
```diff
@@ -23,17 +23,26 @@ export const TEST_SUITE_RESULTS: AxeTestResult[] = [];
...
 let currentTestResult: AxeTestResult | undefined = undefined;
 let currentViolationsMap: Map<string, Result> | undefined = undefined;
-let currentUrls = new Set<string>();
-let currentRouteNames = new Set<string>();
+let currentUrls: Set<string> | undefined;
+let currentRouteNames: Set<string> | undefined;
+let currentTestHelpers: Set<string> | undefined;
 
+/**
+ * A custom reporter that is invoke once per failed a11yAudit call. This can be called
```
drewlee, line 31:
```suggestion
 * A custom reporter that is invoked once per failed a11yAudit call. This can be called
```

drewlee (approved):

addon-test-support/setup-middleware-reporter.ts
```diff
@@ -42,16 +51,23 @@ export async function middlewareReporter(axeResults: AxeResults) {
...
+    currentUrls = new Set();
+    currentRouteNames = new Set();
+    currentTestHelpers = new Set();
   }
 
-  currentUrls.add(currentURL());
-  currentRouteNames.add(currentRouteName());
+  currentTestHelpers = new Set([
+    ...currentTestHelpers!.values(),
+    ...testMetaData.usedHelpers,
```
rwjblue, line 67: If we stash off the `testMetaData` (into `currentTestResult`), we can avoid using `currentTestHelpers` at all here and instead update `pushTestResult` to grab the `usedHelpers` at the end

addon-test-support/setup-middleware-reporter.ts
```diff
@@ -23,17 +23,26 @@ export const TEST_SUITE_RESULTS: AxeTestResult[] = [];
...
+ *
+ * @param axeResults The axe results for each a11yAudit.
+ * @returns Early returns if no violations are found.
+ */
 export async function middlewareReporter(axeResults: AxeResults) {
   if (axeResults.violations.length === 0) {
     return;
   }
 
+  let testMetaData = getTestMetadata(getContext());
```
rwjblue, line 42: We should stash this off into the `currentTestResult`, then in `pushTestResult` use the metadata to grab the `usedHelpers`.

addon-test-support/setup-middleware-reporter.ts
scalvert: Good call. Will change that.

addon-test-support/setup-middleware-reporter.ts
scalvert: 👍

drewlee (approved):
LGTM

PR #270, opened 2021-06-21
Fixes URL mutation when setting the `enableA11yAudit` query parameter

drewlee:
When programmatically setting the `enableA11yAudit` query parameter using the `new URL()` API, the query parameters without an associated value are "normalized".

For example, the following search string
`?hidepassed=true&dockcontainer&split=5&loadBalance&partition=1&browser=10`,
gets mutated to
`?hidepassed=true&dockcontainer=&split=5&loadBalance=&partition=1&browser=10&enableA11yAudit=`,
where an extra `=` is inserted between each key/value pair (including the new `enableA11yAudit` key).

This unintentional side effect may cause issues for other areas of the app that rely on a specific query parameter format, such as in `ember-exam`, where the mutation causes issues in properly determining load balancing. The proposed fix strips out the extraneous characters from the URL `href` before rewriting the window's history. This way, the mutated search string
`?hidepassed=true&dockcontainer=&split=5&loadBalance=&partition=1&browser=10&enableA11yAudit=`,
is transformed back to its original form:
`?hidepassed=true&dockcontainer&split=5&loadBalance&partition=1&browser=10&enableA11yAudit`.

I created a test case to also demonstrate the issue when the fix is not implemented:
Screen Shot 2021-06-21 at 10 06 45 AM

### DIFF
--- a/addon-test-support/index.ts
+++ b/addon-test-support/index.ts
@@ -1,11 +1,7 @@
 export { default as a11yAudit } from './audit';
 export { default as a11yAuditIf } from './audit-if';
 export { setRunOptions, getRunOptions } from './run-options';
-export {
-  setEnableA11yAudit,
-  shouldForceAudit,
-  calculateUpdatedHref as _calculateUpdatedHref,
-} from './should-force-audit';
+export { setEnableA11yAudit, shouldForceAudit } from './should-force-audit';
 export { useMiddlewareReporter } from './use-middleware-reporter';
 export {
   setupGlobalA11yHooks,
--- a/addon-test-support/index.ts
+++ b/addon-test-support/index.ts
@@ -1,7 +1,11 @@
 export { default as a11yAudit } from './audit';
 export { default as a11yAuditIf } from './audit-if';
 export { setRunOptions, getRunOptions } from './run-options';
-export { setEnableA11yAudit, shouldForceAudit } from './should-force-audit';
+export {
+  setEnableA11yAudit,
+  shouldForceAudit,
+  calculateUpdatedHref as _calculateUpdatedHref,
+} from './should-force-audit';
 export { useMiddlewareReporter } from './use-middleware-reporter';
 export {
   setupGlobalA11yHooks,
--- a/addon-test-support/should-force-audit.ts
+++ b/addon-test-support/should-force-audit.ts
@@ -1,7 +1,5 @@
 import { ENABLE_A11Y_AUDIT } from './cli-options';
 
-const A11Y_AUDIT_PARAM = 'enableA11yAudit';
-
 export function _calculateUpdatedHref(
   href: string,
   baseURI: string,
@@ -12,15 +10,15 @@ export function _calculateUpdatedHref(
 
   // Set up the `enableA11yAudit` query param
   if (enabled) {
-    url.searchParams.set(A11Y_AUDIT_PARAM, '');
+    url.searchParams.set('enableA11yAudit', '');
   } else {
-    url.searchParams.delete(A11Y_AUDIT_PARAM);
+    url.searchParams.delete('enableA11yAudit');
   }
 
   // Match all key-only params with '='
   return url.href.replace(/([^?&]+)=(?=&|$)/g, (match, sub) => {
     // Only normalize `enableA11yAudit` or params that didn't initially include '='
-    return sub === A11Y_AUDIT_PARAM || !initialHref.includes(match)
+    return sub === 'enableA11yAudit' || !initialHref.includes(match)
       ? sub
       : match;
   });
--- a/addon-test-support/should-force-audit.ts
+++ b/addon-test-support/should-force-audit.ts
@@ -1,7 +1,11 @@
 import { ENABLE_A11Y_AUDIT } from './cli-options';
 
-export function setEnableA11yAudit(enabled: boolean = false) {
-  const url = new URL(window.location.href, document.baseURI);
+export function calculateUpdatedHref(
+  href: string,
+  baseURI: string,
+  enabled: boolean = false
+): string {
+  const url = new URL(href, baseURI);
 
   // set up the enableA11yAudit query param
   if (enabled) {
@@ -10,8 +14,17 @@ export function setEnableA11yAudit(enabled: boolean = false) {
     url.searchParams.delete('enableA11yAudit');
   }
 
+  return url.href.replace(/=(?=&)|=$/g, '');
+}
+
+export function setEnableA11yAudit(enabled: boolean = false) {
+  const href = calculateUpdatedHref(
+    window.location.href,
+    document.baseURI,
+    enabled
+  );
+
   // updates the URL without reloading
-  const href = url.href.replace(/=(?=&)|=$/g, '');
   window.history.replaceState(null, '', href);
 }
 
--- a/addon-test-support/should-force-audit.ts
+++ b/addon-test-support/should-force-audit.ts
@@ -8,21 +8,7 @@ export function calculateUpdatedHref(
   enabled: boolean = false
 ): string {
   const url = new URL(href, baseURI);
-
-  // Find all query params with no '='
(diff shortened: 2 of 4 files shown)


scalvert:
This seems fine overall (I believe @rwjblue has some comments). We should absolutely fix the url parsing/manipulating in [`ember-exam`](https://github.com/ember-cli/ember-exam/blob/master/addon-test-support/-private/get-url-params.js#L12) to use the correct APIs (URL). Once that's fixed, we could deprecate this hack to keep the internals clean.

rwjblue (changes requested):
Let's split the `setEnableA11yAudit` function up into two functions:

```js
export function _calculateUpdatedHref(href: string, baseURI: string, enabled: boolean = false): string {
  const url = new URL(href, baseURI);

  // set up the enableA11yAudit query param
  if (enabled) {
    url.searchParams.set('enableA11yAudit', '');
  } else {
    url.searchParams.delete('enableA11yAudit');
  }

  const href = url.href.replace(/=(?=&)|=$/g, '');

  return href;
}

export function setEnableA11yAudit(enabled: boolean = false) {
  const href = _calculateUpdatedHref(window.location.href, document.baseURI, enabled);

  // updates the URL without reloading
  window.history.replaceState(null, '', href);
}
```

Then instead of mucking with `window.history` in your test, you can call this new utility function with a large number of permutations of inputs and assert on the expected outputs.

drewlee:
Thanks for the suggestion @rwjblue, that did simplify testing a bit!

tests/acceptance/a11y-audit-test.ts
```diff
@@ -137,4 +138,27 @@ module('Acceptance | a11y audit', function (hooks) {
...
+    );
+
+    href = _calculateUpdatedHref(expectedHref, baseUrl, false);
+    expectedHref = url;
+    assert.equal(
+      href,
+      expectedHref,
+      'Updated URL matches the expected URL when `enableA11yAudit` is removed'
+    );
+  });
```
rwjblue, line 163: We need a few more test cases:

* when `&enableA11yAudit` is already present and you call `_calculateUpdateHref(url, baseUrl, true)` it shouldn't change anything
* when `&enableA11yAudit` is not present and you call `_calculateUpdateHref(url, baseUrl, false)` it shouldn't do anything
* when `&enableA11yAudit=` is present and you call `_calculateUpdateHref(url, baseUrl, true)` it should normalize to remove the trailing `=` 
* when `&enableA11yAudit=` is present and you call `_calculateUpdateHref(url, baseUrl, false)` it should remove the flag appropriately
* does not normalize other unrelated search params (we cannot assume that **everyone** using any possible query string param wants their `=` removed)

rwjblue: I think a small bit of meta-programming will make these test permutations much easier, I'm thinking something like:

```js
  module('enableA11yAudit Query Param', function() {
    const baseUrl = 'https://www.example.com';

    function verify({ label, input, enabled, expected }) {
      test(`${label}: setEnableA11yAudit('${input}', '${baseUrl}', ${enabled}) -> ${output}`, function(assert) {
        let actual = _calculateUpdatedHref(`${baseUrl}/${input}`, baseUrl, expected);

        assert.equal(actual, expected);
      });
    }
    
    verify({
      label: 'can enable',
      input: '/some/path?foo=123&bar&baz',
      enabled: true,
      expected: `/some/path?foo=123&bar&baz&enableA11yAudit`,
    });
    
    verify({
      label: 'can disable',
      input: '/some/path?foo=123&bar&baz&enableA11yAudit',
      enabled: false,
      expected: `/some/path?foo=123&bar&baz`,
    });
    
    verify(
      label: 'already enabled, enabling noops',
      input: '/some/path?foo=123&bar&baz&enableA11yAudit',
      enabled: true,
      expected: `/some/path?foo=123&bar&baz&enableA11yAudit`,
    });
    
    verify({
      label: 'already enabled, enabling normalizes',
      input: '/some/path?foo=123&bar&baz&enableA11yAudit=',
      enabled: true,
      expected: `/some/path?foo=123&bar&baz&enableA11yAudit`,
    });
    
    verify({
      label: 'does not normalize unknown params',
      input: '/some/path?foo=&enableA11yAudit=',
      enabled: true,
      expected: `/some/path?foo=&enableA11yAudit`,
    });
  })
```

rwjblue (changes requested):

tests/acceptance/a11y-audit-test.ts
scalvert: This seems fine to have the iterative test cases. I'd personally prefer not to have a function like this, but instead just loop with an object of case options. Also, we're testing `_calculateUpdatedHref`, not `setEnableA11yAudit`, so I'd probably reflect that in the test case name.

drewlee: I've beefed up the tests and had to change the param logic considerably to handle the case of not normalizing unrelated params. Let me know how this looks.
Screen Shot 2021-06-22 at 8 51 43 AM

addon-test-support/index.ts
```diff
@@ -1,7 +1,11 @@
 export { default as a11yAudit } from './audit';
 export { default as a11yAuditIf } from './audit-if';
 export { setRunOptions, getRunOptions } from './run-options';
-export { setEnableA11yAudit, shouldForceAudit } from './should-force-audit';
+export {
+  setEnableA11yAudit,
+  shouldForceAudit,
+  calculateUpdatedHref as _calculateUpdatedHref,
```
scalvert, line 7: If we're planning on exporting this as public API why do we need to underscore it? Does it need to be public?

addon-test-support/should-force-audit.ts
```diff
@@ -1,17 +1,41 @@
-import { ENABLE_A11Y_AUDIT } from './cli-options';
+import { ENABLE_A11Y_AUDIT as ENABLE_A11Y_AUDIT_ENV } from './cli-options';
```
scalvert, line 1: Not a huge fan of doing this, as it makes it more confusing to read when the name changes to a local alias in certain files. Can we figure out better names so we don't have to alias?

addon-test-support/should-force-audit.ts
```diff
@@ -1,17 +1,41 @@
...
 
-  // updates the URL without reloading
-  window.history.replaceState(null, '', url.href);
+  // Match all key-only params with '='
+  return url.href.replace(/([^?&]+)=(?=&|$)/g, (match, sub) => {
+    // Only normalize `enableA11yAudit` or params that didn't initially include '='
+    if (sub === ENABLE_A11Y_AUDIT || !initialHref.includes(match)) {
+      return sub;
+    }
+    return match;
```
scalvert, line 26: You could probably just return from a ternary here.

scalvert (changes requested):

addon-test-support/index.ts
drewlee: It should be private. I only included it here to make it easier to import from tests (similar to the middleware reporter utils), but I can just import from the module file directly.

addon-test-support/should-force-audit.ts
drewlee: Yeah, no problem.

addon-test-support/should-force-audit.ts
```diff
@@ -1,17 +1,40 @@
 import { ENABLE_A11Y_AUDIT } from './cli-options';
 
-export function setEnableA11yAudit(enabled: boolean = false) {
-  const url = new URL(window.location.href, document.baseURI);
+const A11Y_AUDIT_PARAM = 'enableA11yAudit';
```
scalvert, line 3: We could also just inline this, since it makes it easier to read 🤷‍♂️

scalvert (approved):

addon-test-support/should-force-audit.ts
drewlee: Yeah, there was more usage before I refactored. Thanks for the fresh eyes!

PR #271, opened 2021-06-30
Migrates to native dependabot

scalvert:
Dependabot has become the de facto dependency management bot since its acquisition by Github. Since it's now integrated as part of Github workflows, and to potentially take advantage of any future improvements, this PR standardizes us to use Dependabot vs. Renovate. 

Note: I also noticed we had an ember-try file checked in, so deleted it.


drewlee (approved):
Thanks for updating this!

PR #272, opened 2021-07-01
Fixes issue when violations count is reported to be 0

scalvert:
It's possible when encountering violations that the violations' node array is empty. In that case, when we try to format the violation, we output a 0 count for violations due to the fact we try to infer the violations count occurrences through the presence of node's html.

In addition, the `formatViolation` tests were not indicative of the actual execution of this function. The function signature indicated that both violation and markup could be either undefined or optional respectively, but this did not reflect how this was used in practice. The two parameters are always required, and this misalignment helped manifest this bug.

This PR addresses both fixing the issue, and updating the tests to better reflect the usage.

### DIFF
--- a/addon-test-support/format-violation.ts
+++ b/addon-test-support/format-violation.ts
@@ -3,8 +3,8 @@ import type { Result } from 'axe-core';
 /**
  * Formats the axe violation for human consumption
  *
- * @param {AxeViolation} violation
- * @param {string | string[]} markup (optional) string of HTML relevant to the violation
+ * @param {Partial<Result>} violation
+ * @param {string[]} markup (optional) string of HTML relevant to the violation
  */
 export default function formatViolation(
   violation: Partial<Result>,
--- a/addon-test-support/format-violation.ts
+++ b/addon-test-support/format-violation.ts
@@ -7,14 +7,9 @@ import type { Result } from 'axe-core';
  * @param {string | string[]} markup (optional) string of HTML relevant to the violation
  */
 export default function formatViolation(
-  violation: Partial<Result> | undefined,
-  markup?: string | string[]
+  violation: Partial<Result>,
+  markup: string[]
 ) {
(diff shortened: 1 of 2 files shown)


rwjblue (approved):

addon-test-support/format-violation.ts
```diff
@@ -7,34 +7,25 @@ import type { Result } from 'axe-core';
...
 
-  if (markup) {
-    if (Array.isArray(markup)) {
-      count = markup.length;
-      markup = markup.join('\n');
-    }
-    markup = ` Offending nodes are: \n${markup}`;
-  } else {
-    markup = '';
+  if (markup.length) {
```
ckundo, line 22: is it possible to have a violation without a corresponding node?

ckundo (approved):
looks good, please update the docs and 🚀

drewlee (approved):
LGTM

addon-test-support/format-violation.ts
scalvert: Based on the [type definition in the `axe.Result`](https://github.com/dequelabs/axe-core/blob/42738b5258f058a36b00533e8c5cfdc0f1bcdbed/axe.d.ts#L114) type, `nodes` is always at least an empty array. When we [map over the nodes in the default reporter](https://github.com/ember-a11y/ember-a11y-testing/blob/f142821ab7f22a800eb44c4c5b36970168258146/addon-test-support/reporter.ts#L14), we always at least return an empty array, so this should be safe.

addon-test-support/format-violation.ts
```diff
@@ -7,34 +7,25 @@ import type { Result } from 'axe-core';
  * @param {string | string[]} markup (optional) string of HTML relevant to the violation
```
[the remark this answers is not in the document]
scalvert: 👍

PR #286, opened 2021-07-20
Prevent deletion of a11y audit reports

drewlee:
Current invocations of the middleware reporter cause the deletion of all prior reports from the `ember-a11y-report` directory. This behavior is undesirable as it causes an unexpected removal of prior logs, which may lead to a loss of data if the reports aren't backed up. It also makes it difficult to use with test executions that are spread across multiple partitions, in that only the reports from the last partition get archived. This change removes file deletion so that subsequent reports are additive and includes documentation about the change in behavior.

I think this update merits some discussion on whether it constitutes a major version upgrade. While it isn't necessarily a breaking change, there may be some consumers that depend on auto file deletion and will be unaware that it's something that they may need to explicitly manage going forward.

### DIFF
--- a/README.md
+++ b/README.md
@@ -312,6 +312,8 @@ This addon provides middleware - code that allows the browser to talk to the nod
 
 The middleware reporter writes the results containing all violations detected in all tests to a JSON file stored in a directory, `ember-a11y-report`, in your application or addon's root directory.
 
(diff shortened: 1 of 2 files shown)


drewlee:
Looks like the canary build failure is due to incorrect `deprecate` import in `@ember/test-helpers`: https://github.com/emberjs/ember-test-helpers/pull/1066.

scalvert (approved):
As we discussed offline, this change looks good to me. 

I'm not super opinionated about the major vs. patch discussion, as this _could_ be constituted as a bug, and I don't believe we previously documented the deletion behavior. I would lean towards patch in this case.

rwjblue:
IMHO, this is definitely a bugfix.

PR #288, opened 2021-08-06
Convert to using external middleware reporter package

scalvert:
The middleware reporter functionality contained in this repository had the potential for broader usage than in this repository. Being able to gather data as the result of tests runs and generically store that data can be used for

- `ember-cli-deprecation-workflow`: where we gather all deprecations and write them to a file. This file can then be consumed by external systems to track deprecations.
- validation utilities, where the validations are performed at runtime, during tests, and the results of those validations are needing to be published to metrics systems

In order to support the above use cases, I've extracted this functionality out into a new package: [`ember-setup-middleware-reporter`](https://github.com/scalvert/ember-setup-middleware-reporter). This package contains a collection of utilities that make adding server middleware easier.

This PR updates the custom implementation with the new one from that package.

:warning: Note: systems that rely on the file name pattern defined [here](https://github.com/ember-a11y/ember-a11y-testing/blob/master/setup-middleware.js#L11) should update any dependent logic to support the new format [here](https://github.com/scalvert/ember-setup-middleware-reporter/blob/main/src/index.ts#L126).

Depends on https://github.com/ember-a11y/ember-a11y-testing/pull/289.


drewlee (approved):

drewlee:
>⚠️ Note: systems that rely on the file name pattern defined here should update any dependent logic to support the new format here.
@scalvert would this constitute a breaking change in terms of publishing a new release version?

scalvert:
> >⚠️ Note: systems that rely on the file name pattern defined here should update any dependent logic to support the new format here.
> 
> 
> 
> @scalvert would this constitute a breaking change in terms of publishing a new release version?

@rwjblue and I discussed this, and felt that it's not. You should likely be reading the contents of the specified report directory rather than relying on specific file names.

scalvert:
@dependabot rebase

rwjblue:
>  You should likely be reading the contents of the specified report directory rather than relying on specific file names.

Ya, just to add a bit more color: due to the way we write the files, you can't **really** know ahead of time the precise file name anyways (ember-exam splits, somewhat randomized times from when the file is written, etc). I think in practice, no consumers of these files will have to do updates (because they already had to effectively list the directory and process each file already).

PR #289, opened 2021-08-06
Fixes deprecate call to include for and since props

scalvert:
Fixes the current usages of deprecate to include `for` and `since`. This will ensure CI passes on canary.


drewlee (approved):
LGTM. Thanks for doing this!

PR #291, opened 2021-08-20
Allow @ember/test-waiters 3.x

mydea:
There do not seem to be changes relevant for this addon, so both 2.x and 3.x should be fine:

https://github.com/emberjs/ember-test-waiters/releases/tag/v3.0.0

I also had to update ember-cli-typescript to 4.x, otherwise the embroider build complained (for whatever reason). But as far as I see 4.x also has no relevant changes for this addon: https://github.com/typed-ember/ember-cli-typescript/releases/tag/v4.0.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "postpublish": "ember ts:clean"
   },
   "dependencies": {
-    "@ember/test-waiters": "^2.4.3",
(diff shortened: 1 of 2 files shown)


scalvert:
Thanks, @mydea!

PR #319, opened 2021-11-09
Make helpers configurable in `setupGlobalA11yHooks` mode

jgwhite:
This adds an additional argument, `options`, to `setupGlobalA11yHooks` which allows us to specify additional helpers to hook into.

```js
import {
  setupGlobalA11yHooks,
  DEFAULT_A11Y_TEST_HELPER_NAMES,
} from 'ember-a11y-testing/test-support';

setupGlobalA11yHooks(() => true, {
  helpers: [...DEFAULT_A11Y_TEST_HELPER_NAMES, 'render', 'tab'],
});
```

The motivating use case is to automatically perform an audit when the `render` helper is used, thereby improving automated coverage for component tests.


drewlee:
This would be a breaking change IMO, in that we'd be introducing a new point for a test to fail where one didn't exist before. There are also perf considerations, in that the total volume of a11y audits would substantially increase.

I'm wondering if instead we could make this an opt-in setting, or expose an API to make global hook registration configurable, as this project isn't quite ready for a v5 release. I'm curious to get @scalvert's take, as he authored the hooks implementation, and I'm sure there was a good reason why we didn't leverage `render`.

MelSumner:
@drewlee I think even if we consider this a breaking change AND decide that it should be an option, the option should be to turn it off (having it on should be the default state). WDYT?

scalvert:
I really like this idea for a few reasons:

1. It fills the gap nicely for rendering tests so we have a complement to the application tests that invoke `visit`
2. It provides a tiered set of APIs that cover more test use cases

I also like what you suggested, @MelSumner, WRT to making this an option. What if we, in fact, made all of the hook bindings optional, with specific defaults. This would allow us to add this change, keep the current set of defaults, and give folks a path to using `render` in addition to those defaults. The next major we could then fold `render` into that default set, which would be a natural time for it to graduate.

Thoughts?

tests/integration/components/setup-global-a11y-hooks-test-test.ts
```diff
@@ -0,0 +1,52 @@
+import { module, test } from 'qunit';
```
scalvert, line 1: Can we rename this file? I think `setup-global-a11y-hooks-for-render-test.ts` seems like a reasonable name.

tests/integration/components/setup-global-a11y-hooks-test-test.ts
```diff
@@ -0,0 +1,52 @@
...
+  function (hooks) {
+    setupRenderingTest(hooks);
+
+    let actualAuditInvocationsCount = 0;
+
+    function invokeAll(): boolean {
+      return true;
+    }
+
+    async function a11yAuditFake() {
```
scalvert, line 22: I'd prefer we either convert https://github.com/ember-a11y/ember-a11y-testing/blob/master/tests/acceptance/setup-global-a11y-hooks-test.ts#L36 to an `async` function, or we don't use `async` here and return a resolved promise like the existing test is doing. This is mainly for consistency between these very similar tests.

tests/integration/components/setup-global-a11y-hooks-test-test.ts
```diff
@@ -0,0 +1,52 @@
...
+        <button type="button">Hello!</button>
+      `);
+      await click('button');
+
+      // We expect three invocations because setupRenderingTest performs an
+      // initial render. This is OK, because that initial render is empty and
+      // therefore should not contain any a11y violations.
+      assert.equal(
+        actualAuditInvocationsCount,
+        3,
```
scalvert, line 47: Where's the 3rd invocation coming from?

scalvert (changes requested):

tests/integration/components/setup-global-a11y-hooks-test-test.ts
jgwhite: It's because `setupRenderingTest` calls `render` at the beginning of each test, presumably to set the stage for further rendering. The comment gives a little explanation but let me know if you think it could be clearer.

jgwhite:
Thanks for the splendid feedback and ideas y'all. I'll continue working on it this week.

tests/integration/components/setup-global-a11y-hooks-test-test.ts
jgwhite: We (@gregone and I) figured out a better way to account for this so it’s less confusing now and doesn’t require a comment.

jgwhite:
@gregone and I addressed all the feedback and added the new configuration option in the most pragmatic way possible. Looking forward to hearing what y’all think.

README.md
```diff
@@ -54,16 +54,40 @@ export interface InvocationStrategy {
...
+  | 'visit';
+
+export const defaultA11yHelperNames = ['visit', 'click', 'doubleClick', 'tap'];
+
 export function setupGlobalA11yHooks(
   shouldAudit: InvocationStrategy,
-  audit: (...args: any[]) => PromiseLike<void> = a11yAudit
+  audit: (...args: any[]) => PromiseLike<void> = a11yAudit,
+  options: GlobalA11yHookOptions = { helpers: defaultA11yHelperNames }
 );
```
scalvert, line 83: I'm inclined to change the signature here. The current signature has a default parameter that allows you to invoke it without having to pass the `a11yAudit` helper.

```ts
setupGlobalA11yHooks(() => true);
```

With this change, consumers will be forced to invoke this function by always passing the `a11yAudit` function in when they want to pass `GlobalA11yHookOptions`. I think it would make sense to perhaps move the `audit` param into the `GlobalA11yHooksOptions` options object, which would allow us to pass overrides in, while still providing defaults. We'd want to instead move the defaults merge into the function body, and likely make options optional (hehe). 

It would look something like this:

```ts
export function setupGlobalA11yHooks(
  shouldAudit: InvocationStrategy,
  options?: GlobalA11yHookOptions
);
```

And then inside the function itself we'd want to merge the options with defaults. For the helpers themselves, I think we'd probably not want to merge these directly, but rather if the helpers prop was not supplied in the options, we'd want to set it to the defaults. If it is provided, we'd assume you want to control the whole thing, and have done the associated merge with the default helpers _if you want to use those helpers_. 

I should point out that this change I'm proposing would _also_ precipitate a breaking change, which would require a major version bump. That said, that's what majors are for 🤷‍♂️ .

README.md
```diff
@@ -85,6 +109,20 @@ start();
...
+By default, audits will be run on `visit`, `click`, `doubleClick`, and `tap`. To add additional helpers to hook into, specify them in the options argument:
+
+```js
+import {
+  setupGlobalA11yHooks,
+  a11yAudit,
+  defaultA11yHelperNames,
+} from 'ember-a11y-testing/test-support';
+
+setupGlobalA11yHooks(() => true, a11yAudit, {
```
scalvert, line 121: My comment above would allow us to avoid this.

README.md
```diff
@@ -249,12 +287,12 @@ To do so, import and use `shouldForceAudit` from `ember-a11y-testing`, as shown
...
+    await visit('/');
 
-  if (shouldForceAudit()) {
-    await a11yAudit();
-  }
-  assert.ok(true, 'no a11y errors found!');
+    if (shouldForceAudit()) {
+      await a11yAudit();
+    }
+    assert.ok(true, 'no a11y errors found!');
```
scalvert, line 295: Incorrect indent?

README.md
```diff
@@ -263,12 +301,12 @@ test('Some test case', await function(assert) {
...
-  await visit('/');
+    await visit('/');
 
-  if (shouldForceAudit()) {
+    if (shouldForceAudit()) {
     await a11yAudit();  // will not run
-  }
-  // ...
+    }
+    // ...
```
scalvert, line 309: Incorrect indent?

scalvert (changes requested):

README.md
drewlee: @scalvert What about just overloading the method signature? This way any of the following would be valid:
```
setupGlobalA11yHooks(shouldAudit)
setupGlobalA11yHooks(shouldAudit, audit)
setupGlobalA11yHooks(shouldAudit, audit, options)
setupGlobalA11yHooks(shouldAudit, options)
```
At least for the time being. Then we can merge `audit` into `options` later down the line and be much more strategic about it in terms of messaging and migration.

scalvert: That's an option too, but the need that would precipitate those overloads would be the fact that when trying to widen signatures with multiple optional parameters, the awkwardness is because we don't have an options object to begin with. If we'd had one, we could easily bolt in additional options into that object.

I would also add that the primary reason for my adding the `audit` param was for testing, and secondarily for consumers to be able to supply their own audit implementation that would likely wrap the existing `a11yAudit`. I'm not sure how common the latter is, but the migration from a param to a param in an options object is quite a simple one, so I'm not sure that presents much of a burden to consumers.

I'm not sure it makes sense to me to avoid moving this into an options object just because we want to avoid a major version bump.

jgwhite: While I find both arguments persuasive, I went with overloading `setupGlobalA11yHooks` as @drewlee proposed. Feels like this will make this change easier to land without the additional overhead of announcing a breaking change.

jgwhite: (Note: I think this means we can remove the `breaking` label)

drewlee: @scalvert I'm not so much concerned about this particular change necessitating a major release, but the fact that we decided to deprecate the `a11yAuditIf` function with v5.0.0. We're using it all over the place and don't yet have a plan for replacement. So I don't think that many consumers (if any) are leveraging the custom `audit` invocation, but quite a lot are using `a11yAuditIf` directly.

scalvert: @drewlee makes sense. Specifically, the replacement is to compose the same functionality together and supply that as the `audit` param in `setupGlobalA11yHooks`. If you want to use it directly, you just write your own helper in your own app, as it was always intended for apps to have their own helpers that could be overridden at the app level.

@jgwhite the thing I don't like about this overload is the need to pass in the audit function in the `setupGlobalA11yHooks` call. That said, I can live with this being temporary so long as we circle back and fix it up for 5.0.0. Thanks for doing all this work!

addon-test-support/setup-global-a11y-hooks.ts
```diff
@@ -1,23 +1,75 @@
...
+): void {
+  let audit: AuditFunction = a11yAudit;
+
+  if (typeof auditOrOptions === 'function') {
+    audit = auditOrOptions;
+  } else {
+    options = auditOrOptions;
+  }
+
+  let helpers = options?.helpers || DEFAULT_A11Y_TEST_HELPER_NAMES;
```
scalvert, line 70: If you declared a separate variable at the top of this scope and initialized it to `{}` you wouldn't need this null coalescing check.

```suggestion
  let audit: AuditFunction = a11yAudit;
  let opts: GlobalA11yHookOptions = {};

  if (typeof auditOrOptions === 'function') {
    audit = auditOrOptions;
  } else {
    opts = auditOrOptions;
  }

  let helpers = opts.helpers || DEFAULT_A11Y_TEST_HELPER_NAMES;
```

You'd also avoid having to turn off the `no-redeclare` rule in our eslint config.

scalvert (changes requested):

.eslintrc.js
```diff
@@ -61,6 +61,8 @@ module.exports = {
       rules: {
         'no-unused-vars': 'off',
         '@typescript-eslint/no-unused-vars': 'error',
+        'no-redeclare': 'off',
+        '@typescript-eslint/no-redeclare': ['error'],
```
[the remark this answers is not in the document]
jgwhite: There was actually a different motivation for this change; I should have documented it better. The `no-redeclare` rule doesn't understand the multiple function signatures in TS overrides and incorrectly labels them as a violation. `@typescript-eslint/no-redeclare` gives all the benefits of `no-redeclare` but without that shortcoming.

README.md
jgwhite: @scalvert with the overloading approach you shouldn't need to pass `audit`; `setupGlobalA11yHooks(auditIf, options)` should work just fine.

README.md
```diff
@@ -248,28 +290,34 @@ To do so, import and use `shouldForceAudit` from `ember-a11y-testing`, as shown
 // `&enableA11yAudit` set in the URL
 import { a11yAudit, shouldForceAudit } from 'ember-a11y-testing/test-support';
 
-test('Some test case', await function(assert) {
-  await visit('/');
+test(
+  'Some test case',
+  await function (assert) {
```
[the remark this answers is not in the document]
jgwhite: Yes, dangit! Will revert

.eslintrc.js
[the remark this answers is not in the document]
jgwhite: Docs for the above: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-redeclare.md

README.md
jgwhite: > Thanks for doing all this work!

@scalvert my pleasure! The introduction of the global hooks system was a dream come true, so thank *you* for that!

addon-test-support/setup-global-a11y-hooks.ts
jgwhite: The linter noticed an issue with this. Will address

.eslintrc.js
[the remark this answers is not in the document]
jgwhite: Note also that this configuration only applies in TypeScript files.

README.md
scalvert: > @scalvert with the overloading approach you shouldn't need to pass `audit`; `setupGlobalA11yHooks(auditIf, options)` should work just fine.

Oh derp, you're right. I completely overlooked that. Thanks!

.eslintrc.js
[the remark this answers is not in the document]
scalvert: Ah, sounds good. Thanks for explaining.

scalvert (approved):

PR #322, opened 2021-11-11
Adds get-port to ensure we have an available port

scalvert:
The port for the middleware tests was hardcoded, and when running them locally this port was in use. This caused the tests to fail, and it forced me to debug locally to determine the cause. 

Adding automated port discovery will alleviate this.


drewlee (approved):
LGTM

rwjblue (approved):

PR #325, opened 2021-11-18
Bump ember-auto-import to 2.2.4 and introduce webpack

SergeAstapov:
There is bunch of wins, greatly summarized by @buschtoens in https://twitter.com/buschtoens/status/1449745394594889734

### DIFF
--- a/config/ember-try.js
+++ b/config/ember-try.js
@@ -62,10 +62,6 @@ module.exports = async function () {
       {
         name: 'ember-release',
         npm: {
-          dependencies: {
-            'ember-auto-import': '^2.2.0',
-            webpack: '^5.0.0',
-          },
           devDependencies: {
             'ember-source': await getChannelURL('release'),
           },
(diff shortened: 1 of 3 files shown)


MelSumner (changes requested):
Pinging @drewlee here too

Do we think this should target a new major version branch?

scalvert:
What would this need to be a major for, @MelSumner?

scalvert:
Is it this:

https://github.com/ef4/ember-auto-import/blob/main/docs/upgrade-guide-2.0.md#quick-summary
> addons that upgrade to ember-auto-import >= 2 will only work in apps that have ember-auto-import >= 2, so they should do their own semver major releases when they upgrade

(I'll be honest, this dependency hell really confuses me :))

SergeAstapov:
@scalvert yes, per above link you posted this is a breaking change hence needs a major.
If there is anything else slated for next major - I would be happy to help

scalvert (approved):

PR #326, opened 2021-11-18
Drop Node.js 10 support

SergeAstapov:
#325 requires major release, so good time to drop old Node.js version support.


scalvert:
Thanks, @SergeAstapov.

scalvert (approved):

PR #328, opened 2021-11-19
Update npmignore file

SergeAstapov:
This shrinks the "unpacked size" for a few Kb

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -27,6 +27,7 @@
 /ember-cli-build.js
 /testem.js
 /tests/
+/node-tests/
 /yarn-error.log
 /yarn.lock
 .gitkeep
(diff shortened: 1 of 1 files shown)


.npmignore
```diff
@@ -24,13 +27,12 @@
 /ember-cli-build.js
 /testem.js
 /tests/
-/node-tests/
```
scalvert, line 27: Removing this has triggered some lint errors. Can you pls restore it?

scalvert (changes requested):

.npmignore
scalvert: I added it back.

scalvert (approved):

Issue #316, opened 2021-11-04
Re-exporting types triggers warning


mydea:
In an app using Embroider, I get this build warning:

```
WARNING in ./node_modules/ember-a11y-testing/test-support/index.js 11:0-64
export 'InvocationStrategy' (reexported as 'InvocationStrategy') was not found in './types' (module has no exports)
 @ ./tests/test-helper.ts 8:0-64 11:0-13
 @ ./assets/test.js 690:0-31

WARNING in ./node_modules/ember-a11y-testing/test-support/index.js 11:0-64
export 'A11yAuditReporter' (reexported as 'A11yAuditReporter') was not found in './types' (module has no exports)
 @ ./tests/test-helper.ts 8:0-64 11:0-13
 @ ./assets/test.js 690:0-31
```

I guess it is because it re-exports TS interfaces:

```ts
export { InvocationStrategy, A11yAuditReporter } from './types';
```

Would it make sense to remove these re-exports, or alternatively just define (and export) these interfaces directly in the `index.ts` file?

PR #331, opened 2021-12-01
Properly re-export types

mydea:
This PR moves type-only modules to be `.d.ts` instead of `.ts` and uses `export type` to re-export them, avoiding warnings in Embroider.

Closes https://github.com/ember-a11y/ember-a11y-testing/issues/316


scalvert:
LGTM. Thanks! (Not dissimilar to https://github.com/emberjs/ember-test-waiters/pull/239!)

scalvert (approved):

Issue #338, opened 2021-12-13
Upgrade blueprint to latest for 5.0 release.


scalvert:
Before we release 5.0 we should upgrade the blueprint of this repo to the latest.

SergeAstapov:
@scalvert can we close this one as v5 got released?

scalvert:
Yep!

PR #264, opened 2021-05-20
Fixes failing CI and crypto error

scalvert:
- sets jquery optional feature to false
- fixes crypto issue

PR #341, opened 2021-12-22
Update addon blueprint with ember-cli-update to 4.0.1

SergeAstapov:


SergeAstapov:
Closes #338

.eslintrc.js
```diff
@@ -25,23 +25,17 @@ module.exports = {
     // node files
     {
       files: [
-        '.eslintrc.js',
-        '.template-lintrc.js',
-        'ember-cli-build.js',
-        'index.js',
-        'setup-middleware.js',
-        'cli-options-filter.js',
-        'node-tests/setup-middleware-test.js',
```
scalvert, line 34: Do we not need this anymore?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,80 @@
+name: CI
```
scalvert, line 1: This name change will break our build badge in the README.md. Can we update that?

ember-cli-build.js
```diff
@@ -18,12 +19,10 @@ module.exports = function (defaults) {
 
   const { maybeEmbroider } = require('@embroider/test-setup');
   return maybeEmbroider(app, {
-    packagerOptions: {
-      webpackConfig: {
-        node: {
-          crypto: 'empty',
```
scalvert, line 24: Is this not needed anymore?

package.json
```diff
@@ -46,30 +46,33 @@
...
+    "test": "npm-run-all lint test:*",
     "test:node": "qunit node-tests",
     "test:ember": "ember test",
     "test:ember-compatibility": "ember try:each",
     "prepublishOnly": "ember ts:precompile",
     "postpublish": "ember ts:clean"
   },
   "resolutions": {
-    "socket.io": "^4.3.1"
+    "fuse.js": "6.4.6"
```
scalvert, line 65: Can you explain this?

package.json
```diff
@@ -78,59 +81,58 @@
...
-    "eslint-config-prettier": "^8.0.0",
-    "eslint-plugin-ember": "^10.0.0",
+    "ember-try": "^2.0.0",
+    "eslint": "^7.32.0",
+    "eslint-config-prettier": "^8.3.0",
+    "eslint-plugin-ember": "^10.5.8",
     "eslint-plugin-node": "^11.1.0",
-    "eslint-plugin-prettier": "^3.1.4",
-    "express": "^4.17.1",
+    "eslint-plugin-prettier": "^4.0.0",
```
scalvert, line 121: I wonder if we should remove this in favor of using `prettier --check .`, which is [recommended](https://prettier.io/docs/en/integrating-with-linters.html#notes).

scalvert (changes requested):

scalvert:
@SergeAstapov thanks for doing this, and thanks for reaching out in discord!

.eslintrc.js
SergeAstapov: @scalvert it should be covered by more generic `'./node-tests/**/*.js',` few lines below

.github/workflows/ci.yml
SergeAstapov: @scalvert good point! Will update

ember-cli-build.js
SergeAstapov: @scalvert actually this was giving me hard error:

```
Invalid configuration object. Webpack has been initialized using a configuration object that does not match the API schema.
51
 - configuration[0].node has an unknown property 'crypto'. These properties are valid:
52
   object { __dirname?, __filename?, global? }
53
   -> Options object for node compatibility features.
```

Looking into [PR where it was introduced](https://github.com/ember-a11y/ember-a11y-testing/pull/264) seem like it was needed to fix breaking build. but now it causes failure 🙃

package.json
SergeAstapov: @scalvert got failure for the "Floating Dependencies" job in https://github.com/ember-a11y/ember-a11y-testing/runs/4608355862?check_suite_focus=true with error

```
error fuse.js@6.5.0: The engine "node" is incompatible with this module. Expected version ">=14". Got "12.22.1"
```

apparently fuse.js@6.5.0 had breaking change yesterday, there is logged issue https://github.com/krisk/Fuse/issues/600

SergeAstapov: previous `socket.io` resolutions seem not needed anymore

package.json
SergeAstapov: @scalvert interesting! I think we may do so, this comes from default ember-cli blueprint, so not sure if it worth to diverge here

.github/workflows/ci.yml
SergeAstapov: @scalvert updated!

.eslintrc.js
scalvert: Ah sorry I missed that. Thanks!

.github/workflows/ci.yml
scalvert: Thank you!

ember-cli-build.js
scalvert: Ya, we can remove it and see if it breaks anything :)

package.json
scalvert: Makes sense, and thanks for the explanation.

package.json
scalvert: Ya let's leave it for now, and we can work on updating the default blueprint.

scalvert (approved):

scalvert:
Thanks again, @SergeAstapov!

package.json
SergeAstapov: @scalvert fuse.js issue was resolved and new patch version released. Rised #352 to remove this resolution

PR #344, opened 2021-12-22
Add typechecking compat scenarios to CI

SergeAstapov:
This is inspired by setup in [`ember-modifier`](https://github.com/ember-modifier/ember-modifier/blob/master/config/ember-try-typescript.js)


scalvert (approved):
Nice!

PR #351, opened 2021-12-23
Bump node-fetch to v3 and get-port to v6

SergeAstapov:
As QUnit is not able handle ESM, need to use CJS/ESM interop for imports is node tests.

Similar PRs #347 and #321 by dependabot fail without such change.


node-tests/setup-middleware-test.js
```diff
@@ -42,6 +40,7 @@ QUnit.module('setupMiddleware', function (hooks) {
       urlPath: 'report-violations',
       reportDir: 'ember-a11y-report',
     });
+    let { default: getPort } = await import('get-port');
```
SergeAstapov, line 43: would love to learn a better way of doing this kind of interop if it exists

scalvert: Using `import` is the way to do this type of thing. We could also just not upgrade to these breaking majors, and wait for esm to settle a bit more before converting. Either way works for me, and since this is only needed in tests I think this is fine.

SergeAstapov: @scalvert totally agree that this is not a must at this point of time (and even maybe for another year or two, until old versions stop working, if it happens at all).

It's just something I noticed as was curios of the failure in dependabot PRs and went this rabbit hole.

Feel free to close as this actually may cause more churn that benefit.

scalvert: I'm totally fine with merging. Thanks!

Issue #489, opened 2023-01-24
Test error after upgrading `@ember/test-helpers` to 2.9.0


eleenyeh:
Hello!

I'm starting to see this error consistently in integration tests after upgrading `@ember/test-helpers` to 2.9.0:
```
Assertion Failed: currentRouteName shoudl be a string
```
Looks like the new version adds an [assert](https://github.com/emberjs/ember-test-helpers/blob/master/addon-test-support/%40ember/test-helpers/setup-application-context.ts#L200) on the `currentRouteName` helper which is used by [middleware-reporter](https://github.com/ember-a11y/ember-a11y-testing/blob/3947e4ee18a2c6f73eb49af08397d2d252b1e64d/addon-test-support/setup-middleware-reporter.ts#L76). I think we should avoid calling `currentRouteName` outside of acceptance tests. Let me know what you think! Thanks :)

drewlee:
PR #491 addresses the issue.

drewlee:
Fix has been merged and will be followed up with a release.

PR #491, opened 2023-02-01
Fixes currentRouteName middleware reporter test failures

drewlee:
Fixes issue #489.

The recent release of `@ember/test-helpers` v2.9.0 added an [assertion error](https://github.com/emberjs/ember-test-helpers/blob/ded104ba04e354bfefca4d0b4d2079eb70f29814/addon-test-support/%40ember/test-helpers/setup-application-context.ts#L200) that is emitted if the route name isn't available for a particular test. This results in a slew of unrelated application test failures when running the middleware reporter.

The resulting test failures typically consist of:
```sh
Promise rejected during "clicking the View Permissions button brings up the View modal": Assertion Failed: currentRouteName shoudl be a string"Source:
Error: Assertion Failed: currentRouteName shoudl be a string
    at assert (http://localhost:4444/talent/assets/vendor-static.js:32729:15)
    at currentRouteName (http://localhost:4444/talent/assets/test-support.js:22729:75)
    at middlewareReporter (http://localhost:4444/talent/assets/test-support.js:32566:61)
    at async http://localhost:4444/talent/assets/test-support.js:32499:11
    at async Promise.all (index 0)
    at async Object.<anonymous> (http://localhost:4444/talent/assets/tests.js:32489:7)"
```

The fix introduces a loose wrapper around the `currentRouteName` call, which absorbs the error, preventing it from getting emitted up to the test scope. All other behavior remains unchanged.

### DIFF
--- a/addon-test-support/setup-middleware-reporter.ts
+++ b/addon-test-support/setup-middleware-reporter.ts
@@ -39,6 +39,30 @@ let currentUrls: Set<string> | undefined;
 let currentRouteNames: Set<string> | undefined;
 
 /**
+ * Utility to retrieve the route name corresponding to the current test. Absorbs the emitted
+ * assertion error if route name is `null`, resulting in an empty string return value.
+ *
+ * @param getFn Function to use to derive the route name.
+ * @returns Route name or empty string.
+ */
+export function _getCurrentRouteName(getFn = currentRouteName): string {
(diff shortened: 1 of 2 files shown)


scalvert (approved):
Seems unfortunate that this breaking change caused such a downstream affect. Thanks for fixing!

Issue #498, opened 2023-02-13
TypeScript fails when using latest `@ember/test-helpers`


simonihmig:
The latest version of `@ember/test-helpers` ships its own types, so `@types/ember__test-helpers` is obsolete now (see https://github.com/typed-ember/ember-cli-typescript/pull/1545).

In a project with that setup (i.e. no `@types/ember__test-helpers`) I am getting this when installing ember-a11y-testing:

```
../node_modules/.pnpm/ember-a11y-testing@5.2.0_xxksirw4yhzqkev2rkehjstyum/node_modules/ember-a11y-testing/test-support/setup-middleware-reporter.d.ts:1:23 - error TS2688: Cannot find type definition file for 'ember__test-helpers'
[tests] 1 /// <reference types="ember__test-helpers" />
```

See the `<reference types="ember__test-helpers" />` in the compiled `setup-middleware-reporter.d.ts`, which now references a non-existing file.

NullVoxPopuli:
I think @gitKrystan knows the solution to this type of problem?

simonihmig:
Maybe updating `@ember/test-helpers` and dropping `@types/ember__test-helpers` _here_ would help?

drewlee:
The `registerHook` and `runHooks` functions need to be exposed as official public APIs in `@ember/test-helpers`. This should be fine since they've been in use for several years now. Ember-a11y-testing currently [defines these types locally](https://github.com/ember-a11y/ember-a11y-testing/blob/master/types/%40ember/test-helpers.d.ts.), which unfortunately impedes upgrading `@ember/test-helpers`.

gitKrystan:
This looks to be the same issue I found in ember-modifier (and we'll probably find in more repos as `@types` packages start getting dropped).

> While this project doesn't add the reference directive directly, it appears that with `types: []` ([the default for @tsconfig/ember](https://github.com/tsconfig/bases/pull/149)), the compiler will helpfully(?) add the directive for you. (See https://github.com/microsoft/TypeScript/issues/19874)

There are two possible solutions:

1. Quick but not ideal: https://github.com/ember-modifier/ember-modifier/pull/657
2. Ideal but not quick: https://github.com/ember-modifier/ember-modifier/pull/665 (NOTE: This one isn't ready to merge, so see the discussion in the comments for the Right Way™️.)

gitKrystan:
[...]

PR #503, opened 2023-05-16
Fix for Issue 502

robostheimer:
Change Log:
- removed `ember__test-helpers` from package
- bumped `@ember/test-helpers to v 2.9.2` to get types directly from this package
- Removed HookUnregister import from `setup-global-a11y-hooks` as it is not an exported type and updated the `_unregisterHooks` variable's type to match HookUnregister (Note: happy to make that `HookUnregister` type exportable if that makes more sense)
- Fixed typescript error in setup-middleware-report (`getContext` could return undefine, however `getTestMetadata` expects a defined context)
- removed  `types/@ember/test-helpers.d.ts` since the types are available via the bumped version of @ember/test-helpers

Tests:
`yarn prepublishOnly` returns no errors
`yarn test:ember` returns 
```1..60
# tests 60
# pass  60
# skip  0
# todo  0
# fail  0```

### CODE REVIEW

File: package.json [hunk 1]
```diff
@@ -77,13 +77,12 @@
   },
   "devDependencies": {
     "@ember/optional-features": "^2.0.0",
-    "@ember/test-helpers": "^2.6.0",
+    "@ember/test-helpers": "^2.9.2",
```
chriskrycho, line 80: Let's actually target the latest: 2.9.3 (which has been out for ~5 months or something)!

```suggestion
    "@ember/test-helpers": "^2.9.3",
```

(It won't change the output; Yarn correctly resolved it to that already, but it's best to be explicit.)

addon-test-support/setup-middleware-reporter.ts
```diff
@@ -76,7 +76,8 @@ export async function middlewareReporter(axeResults: AxeResults) {
 
   if (!currentTestResult) {
     let { module, testName } = QUnit.config.current;
-    let testMetaData = getTestMetadata(getContext());
+    const context = getContext() ?? {};
+    let testMetaData = getTestMetadata(context);
```
chriskrycho, line 80: This looks reasonable from a types POV, but that's because while the types in `@ember/test-helpers` in this spot are not *wrong*, they are also not actively preventing us from ending up in a bad spot. If there is no `testContext` here, I expect this may work at a surface level, but that means nothing has correctly called one of `setupTest()`, `setupRenderingTest()`, or `setupApplicationTest()`, all of which call `setupContext()`. Which is… a weird spot to be in!

In the previous version of this, if that ever happened, we would get an error when running the tests, because under the hood, [it's going to try to set up a key-value pair on a `WeakMap`][getTestMetadata]:

[getTestMetadata]: https://github.com/emberjs/ember-test-helpers/blob/c25c5995ec60bc934740aaaf5d85299ede5ba0ee/addon/addon-test-support/@ember/test-helpers/test-metadata.ts#L41-L56

```ts
const TEST_METADATA = new WeakMap<BaseContext, TestMetadata>();

export default function getTestMetadata(context: BaseContext): TestMetadata {
  if (!TEST_METADATA.has(context)) {
    TEST_METADATA.set(context, new TestMetadata());
  }

  return TEST_METADATA.get(context)!;
}
```

If you do that and `getContext()` returns `undefined`, then you will see something like:

> TypeError: WeakMap keys must be objects or non-registered symbols

Net, what we should do here is throw a useful error that tells the end user that they need to make sure to have called `setupTest` etc. first:

```suggestion
    const context = getContext();
    if (!context) {
      throw new Error(
        'You tried to run ember-a11y-testing without calling one of the `setupTest` helpers from `@ember/test-helpers`. Please make sure yoru test setup calls `setupTest()`, `setupRenderingTest()`, or `setupApplicationTest()`!'
      );
    }
    let testMetaData = getTestMetadata(context);
```

chriskrycho (changes requested):
Thanks for doing this. Needs two *very* small tweaks, and then we'll be off to the races!

addon-test-support/setup-middleware-reporter.ts
robostheimer: Sounds reasonable.

package.json
robostheimer: Gotcha, yeah sure.

addon-test-support/setup-global-a11y-hooks.ts
```diff
@@ -24,7 +24,9 @@ type HelperName =
   | 'typeIn'
   | 'visit';
 
-let _unregisterHooks: HookUnregister[] = [];
+let _unregisterHooks: {
+  unregister: () => void;
+}[] = [];
```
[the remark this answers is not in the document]
robostheimer: Sure thing.

chriskrycho (approved):
Excellent, thank you!

chriskrycho:
@rwjblue @scalvert @drewlee this looks good for a 2.9.4 bug-fix release.

rwjblue:
Unfortunately, CI was disabled (unrelated to this PR). I just fixed that.

@robostheimer - Could you rebase / amend & update to trigger a CI job?

addon-test-support/setup-global-a11y-hooks.ts
```diff
@@ -24,6 +24,10 @@ type HelperName =
   | 'typeIn'
   | 'visit';
 
+interface HookUnregister {
+  unregister: () => void;
+}
```
drewlee, line 29: Considering the changes in [PR #1344](https://github.com/emberjs/ember-test-helpers/pull/1344), a 2.9.4 release of @ember/test-helpers would allow us to just import `registerHook` and `HookUnregister` as official public APIs. Is there a reason why we can't just wait for that release before making these changes? Seems like duplicative work.

robostheimer: Right now, I'm trying to migrate an addon to typescript; this PR unblocks that as there is an issue with a reference to ember__test-helpers. Being that the code is already written and ready for merging.  Seems like the duplicative work has been done.

Do we know when the PR referenced above will be released? @drewlee

drewlee: That would be a question for @chriskrycho, who maintains the `@ember/test-helpers` releases.

MelSumner (changes requested):
I agree with @drewlee that it would be better to ask @chriskrycho to please release the package with the fix in it and we can consume that. Sorry for the circles on this, and thank you for your patience! We'll get this resolved. 👍

addon-test-support/setup-global-a11y-hooks.ts
drewlee: To be specific, the duplicative work would entail removing the inline `HookUnregister` type that's being introduced with this change. Just seems more efficient to incorporate these changes wholesale in accommodating a 2.9.4 release of `@ember/test-helpers`. We've been aware of the TypeScript bug (see issue #498 ) and is one of the reasons why the changes were carried out in [PR #1344](https://github.com/emberjs/ember-test-helpers/pull/1344).

robostheimer: I see.  I pushed this PR at the suggestion of @chriskrycho.  Chris, is there a reason why we cant just release `@ember/test-helpers` and then update this addon?

chriskrycho:
One reason is: I'm super busy with other things and keeping these two things decoupled from each other is useful. :) I am happy to try to get to it, but it will be some time before I can commit to that.

robostheimer:
> One reason is: I'm super busy with other things and keeping these two things decoupled from each other is useful. :) I am happy to try to get to it, but it will be some time before I can commit to that.

Gotcha @chriskrycho 

Given this, @drewlee @MelSumner, do you have objections to pushing this change? cc: @rwjblue

MelSumner:
I think we should see who else on the team can release the package; I can even check since as a core framework team member I should have access. Let me check and see.

MelSumner:
> One reason is: I'm super busy with other things and keeping these two things decoupled from each other is useful. :) I am happy to try to get to it, but it will be some time before I can commit to that.

no worries if you don't have time to do a release, I will find someone who can though. I have learned that it's not helpful in the long run to keep things like this limping along and I have the time and energy to see this through. 👍

rwjblue:
We need to address https://github.com/emberjs/ember-test-helpers/pull/1344/files#r1205681209 before releasing a new ember-test-helpers, we should land this as is IMO.

PR #509, opened 2023-06-08
Upgrade @ember/test-helpers to 3.x

### LINKED ISSUE #506: Update ember-test-helpers to v3 latest and cut new release
Since [ember-test-helpers](https://github.com/emberjs/ember-test-helpers/releases/tag/v3.0.0) v3.0 has been released, this addon can be updated and made to work properly. 

New release should be a major and indicate that use of the new version (and going forward) requires use of ETH 3.0 or newer. 

This is probably going to be dependent on ember-qunit also upgrading, though.

fivetanley:
depends upon https://github.com/ember-a11y/ember-a11y-testing/pull/508

partially resolves #506 

- Upgrades `@ember/test-helpers` to `^3.x`.
- Drops support for Node.js < 16.x. These versions are no longer supported by `@ember/test-helpers`. https://github.com/emberjs/ember-test-helpers/blob/92392b110565de201729037b8469362dea1bb472/addon/package.json#L106
- Drop support for Ember < 4

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,15 +7,15 @@ on:
       - master
       - 'v*'
   pull_request: {}
-  workflow_dispatch: 
+  workflow_dispatch:
 
 concurrency:
-   group: ci-${{ github.head_ref || github.ref }}
-   cancel-in-progress: true
+  group: ci-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
 
 jobs:
   test:
-    name: "Tests"
+    name: 'Tests'
     runs-on: ubuntu-latest
 
     steps:
@@ -32,7 +32,7 @@ jobs:
         run: yarn test:node
 
   floating:
-    name: "Floating Dependencies"
+    name: 'Floating Dependencies'
     runs-on: ubuntu-latest
 
     steps:
@@ -55,17 +55,11 @@ jobs:
       fail-fast: false
       matrix:
         try-scenario:
-          - ember-lts-3.8
-          - ember-lts-3.12
-          - ember-lts-3.16
-          - ember-lts-3.20
-          - ember-lts-3.24
-          - ember-lts-3.28
+          - ember-lts-4.4
+          - ember-lts-4.8
           - ember-release
           - ember-beta
           - ember-canary
-          - ember-classic
-          - ember-default-with-jquery
           - embroider-safe
           - embroider-optimized
 
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ individual tests using an `a11yAudit()` test helper.
 - Ember.js v4.0.0 or above
 - Ember CLI v3.8 or above
 - Node.js v16 or above
-- `@ember/test-helpers` v3.0.0 or above
+- `@ember/test-helpers` v3.0.3 or above
 
 Note: we enforce a peerDependency of `@ember/test-helpers`. If you encounter the following message:
 
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ individual tests using an `a11yAudit()` test helper.
 
 ## Compatibility
 
-- Ember.js v3.8.0 or above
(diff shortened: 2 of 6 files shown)


fivetanley:
It looks like this is blocked on `ember-qunit`: https://github.com/ember-a11y/ember-a11y-testing/actions/runs/5214440386/jobs/9410705701?pr=509#step:6:125

We'll need https://github.com/emberjs/ember-qunit/pull/1077 and a new release of ember-qunit first (as @MelSumner pointed out in a slack conversation with me at work).

> I think there might be some changes to make related to what was merged in https://github.com/ember-a11y/ember-a11y-testing/pull/503

It looks like the first time I made the commit I used a newer yarn which significantly changed the lock file. I reverted back to the version in `package.json -> volta -> yarn` and I think we're where we need to be before upgrading ember-qunit.

addon-test-support/setup-global-a11y-hooks.ts
```diff
@@ -1,4 +1,4 @@
-import { _registerHook } from '@ember/test-helpers';
+import { registerHook } from '@ember/test-helpers';
```
drewlee, line 1: We should also import `HookUnregister` directly instead of defining it inline further below.
```suggestion
import { registerHook, type HookUnregister } from '@ember/test-helpers';
```

drewlee (changes requested):

MelSumner:
The README should be updated as part of this PR to tell users what version of the addon to use if they need to support Node 14.

package.json
```diff
@@ -131,7 +131,7 @@
     "webpack": "^5.65.0"
   },
   "peerDependencies": {
-    "@ember/test-helpers": "^2.0.0",
+    "@ember/test-helpers": "^3.0.0",
```
MelSumner, line 134:
```suggestion
    "@ember/test-helpers": "^3.0.3",
```
Only suggesting this change because there were some shenanigans with a few early versions of 3.0 and we're in the process of unpublishing/deprecating them ><

README.md
```diff
@@ -115,7 +117,7 @@ start();
 
 :warning: It's important to note that you must also use the [`enableA11yAudit`](#force-running-audits) query parameter in order to force audits. This setting is required in addition to any invocation strategy you provide.
 
-By default, audits will be run on `visit`, `click`, `doubleClick`, and `tap`. To add additional helpers to hook into, specify them by name in the `options.helpers` argument. Note that this option specifies the *complete* set of helpers to hook into; to include the defaults you must import them and splat them into the array as shown below.
+By default, audits will be run on `visit`, `click`, `doubleClick`, and `tap`. To add additional helpers to hook into, specify them by name in the `options.helpers` argument. Note that this option specifies the _complete_ set of helpers to hook into; to include the defaults you must import them and splat them into the array as shown below.
```
fivetanley, line 120: Most of these changes were unintentional but ran since I have the prettier plugin enabled in VSCode. Should we leave the changes in and figure out how to have prettier check `.md` files in CI in a different PR?

package.json
fivetanley: resolved!

addon-test-support/setup-global-a11y-hooks.ts
fivetanley: resolved!

.github/workflows/ci.yml
```diff
@@ -32,7 +32,7 @@ jobs:
         run: yarn test:node
 
   floating:
-    name: "Floating Dependencies"
+    name: 'Floating Dependencies'
```
fivetanley, line 35: this is a change from prettier

README.md
MelSumner: Yes, I think it's fine to leave them in, can you create an issue to track this though?

MelSumner: #510

MelSumner (approved):

MelSumner:
I'm going to try a pre-release for this so we can see how it works

Issue #505, opened 2023-05-25
Proposal to disable animations and transitions while running a11y tests


tgvrssanthosh:
With animations and transitions in place, the a11y tests can non-deterministically fail.

For eg., lets say we have a disabled button and on certain action (click of another button) we enable the button. Lets say we have a transition property set on the button to transition the background-color of the button.

Now, lets say the disabled button contrast ratio doesn't meet the AA guidelines. The a11y tests will continue to pass since the axe-core doesn't run the tests on disabled elements. In the example that i mentioned above, when the axe-core tests run as soon as the button gets enabled, it can lead to a11y violation error since the background transition might not have completed yet.

I would like to propose to disable animations and transitions while running the a11y tests. We can add the following styles to disable transitions/animations in `test-head-footer` file that already contain few style overrides.

```
.axe-running * {
  animation: none !important;
  transition: none !important;
}
```

Or we can set transition and animation durations to `0` while running the a11y tests as below.
```
.axe-running * {
  animation-duration: none !important;
  transition-duration: none !important;
}
```

drewlee:
I'm fine with this addition since it's something we end up adding as overrides in host applications anyway. Would be good to get @MelSumner's input as well from the wider open source Ember community.

tgvrssanthosh:
@MelSumner Can you please let me know if you are fine with proposal? If so, I can create a PR for the same.

drewlee:
Completed via PR #511.

PR #511, opened 2023-06-17
Disabling transitions and animations while running axe tests

tgvrssanthosh:
Copying the proposal that is added in #505 

With animations and transitions in place, the a11y tests can non-deterministically fail.

For eg., lets say we have a disabled button and on certain action (click of another button) we enable the button. Lets say we have a transition property set on the button to transition the background-color of the button.

Now, lets say the disabled button contrast ratio doesn't meet the AA guidelines. The a11y tests will continue to pass since the axe-core doesn't run the tests on disabled elements. In the example that i mentioned above, when the axe-core tests run as soon as the button gets enabled, it can lead to a11y violation error since the background transition might not have completed yet.

In this PR, we are disabling animation and transition so that the transitions/animations will not affect any of the axe tests
```
.axe-running * {
  animation: none !important;
  transition: none !important;
}
```

### DIFF
--- a/content-for/test-head-footer.html
+++ b/content-for/test-head-footer.html
@@ -28,4 +28,13 @@
     height: 100%;
     transform: scale(1);
   }
+
+  /*
+   * Setting duration to zero so that the elements will get transitioned/animated immediately
+   * so that the gradual transition/animation will not violate any of the WCAG guidelines
+   */
+  .axe-running * {
+    transition-duration: 0s !important;
+    animation-duration: 0s !important;
+  }
 </style>
--- a/content-for/test-head-footer.html
+++ b/content-for/test-head-footer.html
@@ -30,8 +30,7 @@
   }
 
   /*
-   * Disabling animation and transition so that the elements are not evaluated
-   * while the transition/animation is in progress.
+   * Disables visual effects so that elements aren't evaluated while in animation or transition.
    */
   .axe-running * {
     animation: none !important;
--- a/content-for/test-head-footer.html
+++ b/content-for/test-head-footer.html
@@ -30,11 +30,11 @@
   }
 
   /*
-   * Setting duration to zero so that the elements will get transitioned/animated immediately
-   * so that the gradual transition/animation will not violate any of the WCAG guidelines
+   * Disabling animation and transition so that the elements are not evaluated
+   * while the transition/animation is in progress.
    */
   .axe-running * {
-    transition-duration: 0s !important;
-    animation-duration: 0s !important;
+    animation: none !important;
+    transition: none !important;
   }
 </style>


content-for/test-head-footer.html
```diff
@@ -28,4 +28,13 @@
     height: 100%;
     transform: scale(1);
   }
+
+  /*
+   * Setting duration to zero so that the elements will get transitioned/animated immediately
+   * so that the gradual transition/animation will not violate any of the WCAG guidelines
+   */
```
drewlee, line 35: Suggestion to make this comment more concise:
```suggestion
  /*
   * Set duration to zero so that elements aren't evaluated while in transition
   */
```

content-for/test-head-footer.html
```diff
@@ -28,4 +28,13 @@
...
+  .axe-running * {
+    transition-duration: 0s !important;
+    animation-duration: 0s !important;
```
drewlee, line 38: I'd slightly prefer disabling animations and transitions entirely via `animation: none !important; transition: none !important;` as it's been field tested as reliable in quite a bit of internal apps.

I think the main issue with just disabling the duration is that an element may still require some time to switch to the intended state if using `transition-delay`. This uncertainty goes away if we just disable all constituent properties.

content-for/test-head-footer.html
tgvrssanthosh: It's clear and compact now. Thank you!

content-for/test-head-footer.html
tgvrssanthosh: I don't have strong preference one over the other. However, as you mentioned, since disabling of animation and transition approach is battle tested, I switched to that approach.

drewlee:
This change looks good. I'd like to cut a new major 6.0.0 release before merging this change so that it's disassociated from the TypeScript-related work rolled up into v6.

content-for/test-head-footer.html
```diff
@@ -28,4 +28,13 @@
     height: 100%;
     transform: scale(1);
   }
+
+  /*
+   * Disabling animation and transition so that the elements are not evaluated
+   * while the transition/animation is in progress.
```
drewlee, line 34:
```suggestion
   * Disables visual effects so that elements aren't evaluated while in animation or transition.
```

drewlee (approved):

void-mAlex:
this is fine to do in app code and even to provide an opt in to have it be done for you, but as is, the change in this pr breaks any app that uses animation/transition events
tests that were passing just fine without it now fail

I would propose that a different way to handle running a11y tests after animation is to hook into the events associated with them animation(start/end) 
or alternatively reduce the timing to 0 so events still fire but I can imagine even that breaking people's tests that are timing dependant (intentionally or otherwise)

PR #515, opened 2023-09-25
Allow wrapping the default reporter without having to rebuild it

acorncom:
We currently expose `_TEST_SUITE_RESULTS` and `_middlewareReporter` to allow customizing reporting without needing to build the entire set of reporting / middleware / test violation capture. But we don't currently allow wrapping the default reporter which hampers any ability to filter violation results (or report on them differently) while still using the core setup within default reporter.

It's probably worth thinking through ways to expose all of these items as "wrappable" more easily, but short-term I'm working on an app that could use the ability to filter results in certain scenarios


MelSumner (approved):
Let's ship it.

PR #558, opened 2024-10-30
bugfix: prefer using Testem.afterTests instead of QUnit.done

fivetanley:
This affects folks using `setupMiddlewareReporter`.

Sometimes, like in CI environments, there can exist a race condition where Testem closes before the local middleware reporter can finish writing the reports to the filesystem.

According to [Testem's README][1], it appears the proper way to do this is to use `Testem.afterTests` to send reports. This will keep testem open and prevent the process from exiting before the reports are written.

This seems to be a [known issue with Testem's QUnit integration][2].

[1]: https://github.com/testem/testem/blob/0bd05cec7f83bfe9f2e10ddfab54556d724ee179/README.md#running-browser-code-after-tests-complete
[2]: https://github.com/testem/testem/issues/1577

### DIFF
--- a/addon-test-support/setup-middleware-reporter.ts
+++ b/addon-test-support/setup-middleware-reporter.ts
@@ -139,6 +139,16 @@ export function pushTestResult() {
   }
 }
 
+type TestemCallback = (config: any, data: any, callback: () => void) => void;
+
+declare global {
+  interface Window {
+    Testem?: {
+      afterTests: (callback: TestemCallback) => void;
+    };
+  }
+}
+
 /**
  * Sets up the middleware reporter, which reports results when the test suite is done.
  */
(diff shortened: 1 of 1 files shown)


fivetanley:
needs https://github.com/ember-a11y/ember-a11y-testing/pull/559 for typescript CI failure checks

MelSumner:
@fivetanley thanks for the PR. I approved the first PR; I'll review this one and do a release in the morning. 👍

addon-test-support/setup-middleware-reporter.ts
```diff
@@ -149,15 +159,28 @@ export function setupMiddlewareReporter() {
...
+  if (window.Testem) {
+    window.Testem.afterTests(async function (_config, _data, callback) {
+      try {
+        await sendViolationsToServer();
+      } finally {
+        callback();
+      }
+    });
+  } else {
+    QUnit.done(async function () {
```
fivetanley, line 171: falling back to QUnit.done in the rare case that folks are running in an environment other than Testem. I remembered some teams save their runners on S3 or other file stores for debugging purposes.

josephdsumner: The fallback here is appreciated as a review and consumer!

josephdsumner (approved):

josephdsumner (approved):

MelSumner (approved):
Thanks for figuring this out! 

Noting for myself and other reviewers/users/watchers that we'll need to figure out why the `ember-beta` scenario is failing rather soon. 👀

PR #561, opened 2024-10-31
Fixes for `ember-beta` scenario failures

MelSumner:
If merged, this should address the current failing tests in the `ember-beta` test scenario. 

There's still one test that is failing (advice welcome), so I'll keep trying to figure that out. I think something with how the test is set up, or how the test components are set up, is causing this failure to be triggered. But, looking at the associated test, there are several "skip these failures" elements but one still gets through. 🤔 
CleanShot 2024-10-31 at 11 28 06

### DIFF
--- a/tests/dummy/app/templates/violations.hbs
+++ b/tests/dummy/app/templates/violations.hbs
@@ -41,6 +41,4 @@
 
 
 <footer class="p-violations__footer">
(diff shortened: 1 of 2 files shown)


MelSumner:
Fixed in a different PR.

PR #562, opened 2024-10-31
Fix sass deprecations

MelSumner:
Since sass no longer supports `@import`, needed to switch it up to use `@use` instead, which also meant adding a few additional use declarations in files that were pulling in the variables. 

Going to try to find someone who regularly uses sass to review since I haven't used sass in this way in a long time and I'm not entirely positive it's the correct change.

### DIFF
--- a/tests/dummy/app/styles/app.scss
+++ b/tests/dummy/app/styles/app.scss
@@ -1,10 +1,10 @@
-@import 'variables/variables';
+@use 'variables/_variables' as *;
 
-@import 'components/components';
-@import 'object-patterns/object-patterns';
-@import 'pages/pages';
-@import 'typography/typography';
-@import 'utilities/utilities';
+@use 'components/components';
+@use 'object-patterns/object-patterns';
+@use 'pages/pages';
+@use 'typography/typography';
+@use 'utilities/utilities';
 
 html {
   font-size: 100%;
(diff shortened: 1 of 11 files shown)


tests/dummy/app/styles/app.scss
```diff
@@ -1,10 +1,10 @@
-@import 'variables/variables';
+@use 'variables/_variables' as *;
```
didoo, line 1: [question] I don't _think_ the underscore is needed anymore (to have Sass ignore certain files). Is it some form of convention used locally? (I see the other files are "imported/used" without the underscore.

tests/dummy/app/styles/app.scss
```diff
@@ -1,10 +1,10 @@
...
-@import 'pages/pages';
-@import 'typography/typography';
-@import 'utilities/utilities';
+@use 'components/components';
+@use 'object-patterns/object-patterns';
+@use 'pages/pages';
+@use 'typography/typography';
+@use 'utilities/utilities';
 
 html {
```
didoo, line 9: [suggestion] not related to this PR, but I would suggest to move the global declarations in a dedicated file (eg. `global.scss` and import (`@use`) it as first item after the variables, before the other files (components, patterns, etc).

didoo (approved):
I don't see anything strange, or duplicated code in the generated CSS, so for me it's an OK change (based on my very limited knowledge of the codebase).

tests/dummy/app/styles/app.scss
MelSumner: Thank you for the suggestion! I've captured it in a separate issue: https://github.com/ember-a11y/ember-a11y-testing/issues/563

PR #566, opened 2024-11-07
Merge `RunOptions` when calling `a11yAudit`.

deanmarano:
In an app that is disabling rules globally (through `setRunOptions`) and individually (through `a11yAudit`) currently the latter overrides the former.

This PR merges the options, letting the runtime call take precedence, but still allowing global options to pass through. I tested this patch in my host app, and rules were disabled as expected.

### DIFF
--- a/addon-test-support/audit.ts
+++ b/addon-test-support/audit.ts
@@ -58,8 +58,11 @@ export function _normalizeRunParams(
     options = runOptions;
   }
 
-  if (typeof options !== 'object') {
-    options = getRunOptions() || {};
+  let globalRunOptions = getRunOptions() || {};
+  if (typeof options === 'object') {
+    options = { ...options, ...globalRunOptions };
+  } else {
+    options = globalRunOptions;
   }
 
   return [context, options];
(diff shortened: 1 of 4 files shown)


addon-test-support/audit.ts
```diff
@@ -58,9 +58,7 @@ export function _normalizeRunParams(
     options = runOptions;
   }
 
-  if (typeof options !== 'object') {
-    options = getRunOptions() || {};
-  }
+  options = { ...options, ...getRunOptions() };
```
deanmarano, line 61: TIL
```js
{...undefined, ...null, ...{}}
=> {}
```

tests/acceptance/a11y-audit-test.ts
```diff
@@ -137,4 +137,21 @@ module('Acceptance | a11y audit', function (hooks) {
...
+        // Disabled to test whether the config is
+        // properly loaded in test environment
+        'image-alt': { enabled: false },
+      },
+    });
+
+    // Pass unrelated rules at runtime, should not override setRunOptions
+    await a11yAudit({ rules: { 'color-contrast': { enabled: false } } });
+
+    assert.ok(true, 'the image-alt rule should be ignored');
```
MelSumner, line 155: Should you also assert that the color contrast is ignored?

PR #568, opened 2024-11-12
 enable embroider tests in ci

void-mAlex:
moved ember-cli-htmlbars to dependencies to bring it in line with new addon output
https://github.com/ember-cli/ember-addon-output/blob/master/package.json#L33

upgraded e-a-i to v2.10.0 which should unlock all the latest features and js syntax
upgraded @embroider/test-setup to v4 in order to use current stable version of embroider of v3


SergeAstapov:
> moved ember-cli-htmlbars to dependencies

@void-mAlex as addon does not ship any .hbs files, this is technically not needed? Does not hurt for sure, just not needed

SergeAstapov (approved):

void-mAlex:
> > moved ember-cli-htmlbars to dependencies
> 
> @void-mAlex as addon does not ship any .hbs files, this is technically not needed? Does not hurt for sure, just not needed

it came up as an error during embroider build due to some of the other addons present; believe it was ember test helpers. so technically I think it might be needed for the wrong reasons. hope that answer is sufficient

MelSumner (approved):
Thank you for doing this!

PR #571, opened 2024-11-20
Drop ember-destroyable-polyfill that is no longer needed

acorncom:
Per the readme at https://github.com/ember-polyfills/ember-destroyable-polyfill the polyfill is inert after Ember 3.22 Seeing as we now support Ember 4+ only, we can take this out


MelSumner (approved):
Thank you for working on this.

PR #572, opened 2024-11-20
Remove unused body-parser dependency

acorncom:
The body-parser dependency was added in https://github.com/ember-a11y/ember-a11y-testing/commit/cf0d1227c047725d3682101d3c6469145c8a3fdb#diff-2b7b86d547afb814ffca762b5a1162f99ed0f3fbf12e23bd59ed82343d6270e4 but later moved into Steve's `@scalvert/ember-setup-middleware-reporter` add-on. As such, we don't need to keep it up-to-date here


MelSumner (approved):
Good catch

PR #580, opened 2025-02-04
expand test-helpers peer range

mansona:


MelSumner (approved):
(question, non-blocker) do we need to leave this wide of a range in here?

mansona:
So things get a bit funky if you have too many copies of test-helpers (and test-waiters) floating around. If you have a narrow range then the package manager will duplicate things a bit more so keeping this wide a range is good to help deduplicate 👍

PR #597, opened 2025-04-09
[internal] move from yarn to pnpm

mansona:
If merged, this PR switches from yarn to pnpm to align with other upcoming changes.


tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es2017",
+    "target": "es2018",
```
mansona, line 3: I needed to change this because without it I would get this error: 

```
addon-test-support/audit.ts:89:6 - error TS2550: Property 'finally' does not exist on type 'Promise<void>'. Do you need to change your target library? Try changing the 'lib' compiler option to 'es2018' or later.
```

I don't know what was making this pass before 🤷

package.json
```diff
@@ -63,6 +63,7 @@
   },
   "dependencies": {
     "@ember/test-waiters": "^2.4.3 || ^3.0.0 || ^4.0.0",
+    "@glimmer/env": "^0.1.7",
```
mansona, line 66: Yet another one in the category of "I don't know how this was working before" 🙈 

Without this dependency declared typescript would complain, but having it as a devDependency you get this error: 

```
ember-a11y-testing tried to import "@glimmer/env" in "ember-a11y-testing/test-support/setup-middleware-reporter.js" from addon code, but "@glimmer/env" is a devDependency. You may need to move it into dependencies.
```

Which is 100% right 🤷 If you're importing a thing in your code then it needs to go in the dependencies. This is another one of those things that `yarn` hides from you because it doesn't setup your deps right 🙈

MelSumner (approved):
Thank you for doing this!

Issue #595, opened 2025-04-09
Should this repo update package manager and release process?


MelSumner:
We've worked to move a lot of community maintained packages to use `pnpm` at the least, and `release-plan` (where feasible). 

It's probably worth evaluating if this move would be valuable for current maintainers.

PR #598, opened 2025-04-09
[internal] set up release-plan

mansona:
this is just committing the output from `pnpm create release-plan-setup@latest` 👍 it doesn't usually need to be reviewed line-by-line because any issues with it get fed back into the setup system https://github.com/embroider-build/create-release-plan-setup

Once this is merged you will need to setup a NPM_TOKEN before doing the first release https://github.com/embroider-build/create-release-plan-setup?tab=readme-ov-file#create-a-npm_token


MelSumner (approved):
Resolves #595

PR #601, opened 2025-04-09
convert render tests tests to gts

mansona:
I just ran the codemod and it converted everything fine so 🎉


MelSumner (approved):
Thank you for doing this!

PR #659, opened 2026-05-26
Update pnpm install commands in CI workflow

MelSumner:
If merged, this PR updates the github `ci.yml` file. It changes some of the lockfile tags for `pnpm install`; it seems to be causing issues when dependabot wants to update a dependency.


josephdsumner (approved):
Looks great!

